Issue #64, opened 2026-01-19
When updating account balance, the previous balance must be prefill


victor-falcon:
<img width="522" height="370" alt="Image" src="https://github.com/user-attachments/assets/5b02f1c7-4122-4727-aa48-c64a2c9e0277" />

Instead of showing an empty amount on the field, it will be better to show the last amount.

It will be amazing too if we are able to do operations, for example, imagine that the current account balance is $1,000 and I just did a new transaction of +$500.

It will be awesome if, when I open the modal the current amount is already there and I can just type ` + 500`, and press enter.

PR #65, opened 2026-01-19
feat: Better, easier, and faster account balance update modal

victor-falcon:
https://github.com/user-attachments/assets/a80966de-b282-4ba0-9431-2cfbcfddc023

Every detail is important. Now, when you are about to update the balance of an account, the current balance is pre-filled, and you can even do simple math operations.

If you just receive 1,000 dollars, you can just type `+ 1000` and press enter.

No need to do the maths by your self.


victor-falcon:
Fix #64

Issue #391, opened 2026-05-13
docker-compose.production.yml pulls dev image from GHCR — Could not open input file: artisan on plain Docker hosts


raptr55:
### Environment

- Host: OpenMediaVault 7 (Debian 12) with the official Compose plugin
- Docker Engine: 28.x
- Docker Compose: v2.x
- Image pulled: `ghcr.io/whisper-money/whisper-money:latest` (digest of latest `main`)
- No bind mounts other than the `whisper-storage` named volume

### Steps to reproduce

1. On a host with only Docker + Docker Compose installed (no Node, PHP, `whispermoney` CLI, no source clone).
2. Download `docker-compose.production.yml` and `.env.production.example` from `main`.
3. Copy `.env.production.example` to `.env`, set `APP_PORT`, `APP_URL`, DB credentials.
4. `docker compose -f docker-compose.production.yml pull`
5. `docker compose -f docker-compose.production.yml up -d`
6. `docker compose -f docker-compose.production.yml logs -f app`

### Expected behavior

Following the README *"Running with Docker (Production Image)"* should produce a fully working stack on any Docker host, with no host-side build, no `whispermoney` CLI, no source clone — exactly the experience users get with projects like Immich, Vaultwarden, Paperless-ngx, etc.

### Actual behavior

```
mysql-1  | ... ready for connections.
app-1    | Could not open input file: artisan
app-1    | Could not open input file: artisan
app-1    | Could not open input file: artisan
...
```

`docker inspect` confirms the wrong image content:

```
Image=ghcr.io/whisper-money/whisper-money:latest
Entrypoint=[docker-php-entrypoint]
Cmd=[php artisan serve --host=0.0.0.0 --port=8000]
WorkDir=/app
```

`Cmd` is `php artisan serve` (dev) instead of `["/entrypoint.sh"]` (production), and `/app` does not contain `artisan`.

### Additional context

Following the README section *"Running with Docker (Production Image)"* on a plain Docker / Docker Compose host (no Coolify, no `whispermoney` CLI), the `app` container ends up in a restart loop printing:

```
Could not open input file: artisan
```

[...]

PR #393, opened 2026-05-13
  fix: publish and use production Docker image

mouadja02:
## Summary
  - publish a GHCR production image built from `Dockerfile.production`
  - point production Docker Compose and Coolify template to `:production`
  - add `.dockerignore` so host `.env`, `vendor`, and `node_modules` do not leak into image builds
  - clarify Docker self-hosting docs and Compose log command

  ## Verification
  - `docker build -f Dockerfile.production -t whisper-money:production-test .`
  - `docker compose -f docker-compose.production.yml config --images`
  - `vendor/bin/pint --dirty --format agent`
  - `vendor/bin/pest tests/Unit/DockerProductionImageTest.php`


victor-falcon:
Hey, thanks for this! Two things:

I'd like to keep `:latest` as the default in `docker-compose.production.yml` and the Coolify template, that's what my prod server pulls and I don't want to flip the default on existing installs.

Mind reverting just those bits (and the matching test lines)?

The rest (CI publishing `:production`, the `.dockerignore`) is great, happy to merge once that's out.

Also curious, **what's the actual motivation here?** Hitting a real issue self-hosting or running locally? Just want to make sure we're solving the right thing before locking the shape in 🙂

mouadja02:
Hello @victor-falcon 

Done! Reverted docker-compose.production.yml and the Coolify template back to :latest as the default, and updated the test assertions to match.

As for the motivation — issue #391: users self-hosting with docker compose -f docker-compose.production.yml up -d were hitting a pull failure

victor-falcon (approved):

victor-falcon:
I hope it works now. Thanks!

PR #528, opened 2026-06-13
fix: start the new transaction form with no account preselected

tonigruni:
## Summary
When the add/edit transaction dialog resets, it no longer auto-selects the first transactional account — the account field starts empty so the user must pick one deliberately (avoids accidentally booking a transaction to the wrong account).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


victor-falcon:
Since there’s been no reply, I’m closing the pull request.

I’ll open a new one for the same thing. Thanks!

victor-falcon:
https://github.com/whisper-money/whisper-money/pull/549

PR #548, opened 2026-06-17
feat: add personal subscription support

nuageeee:
Hello there,

Here is the base of the personal subscription function.
Need to make the types and obviously the page, but i dont want to pull something unfinished


victor-falcon:
I have a lot of questions.

What’s your idea? Should the user manually register their subscriptions, or how do you plan to detect them?

Also, I think adding a new item to the navigation menu isn’t the right way to do it. Right now, on medium-sized phones the menu already looks pretty cramped.

We need to thing about this.

nuageeee:
In my view, we could add a "Subscription" category. Following that, Whisper would create a card with the name, amount, and date.

So it would be automatic; no need to do it manually unless a subscription isn't deducted from connected accounts.

What do you think about making the "Budgets" category broader? In that case, there's no need to create a new page, but rather, for example, a double page for the same item (I can show you an example if you don't understand).

PR #567, opened 2026-06-19
feat(currency): add RSD (Serbian Dinar)

nvajagic:
## What
Add Serbian Dinar (RSD) as a supported currency.

## Compatibility
Conversion provider `@fawazahmed0/currency-api` covers RSD (standard ISO 4217). Conversion service lowercases codes and fetches `rsd.min.json` - same path NZD uses. `exchange_rates` table stores rates as JSON, so any 3-letter code works. Validation rules and Inertia currency props auto-derive from config via `CurrencyOptions`.

## Changes
- `config/currencies.php` - RSD entry (`allows_primary` + `allows_account`)
- `lang/es.json` - Spanish translation
- `lang/fr.json` - French translation

## Tests
Haven't tested them locally, but should pass

### DIFF
--- a/config/currencies.php
+++ b/config/currencies.php
@@ -146,5 +146,11 @@ return [
             'allows_primary' => false,
             'allows_account' => true,
         ],
+        [
+            'code' => 'RSD',
+            'name' => 'Serbian Dinar',
+            'allows_primary' => true,
+            'allows_account' => true,
+        ],
     ],
 ];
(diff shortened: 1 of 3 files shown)


victor-falcon:
Looks good, let's merge it when CI is fully green :). Thanks

victor-falcon:
It should be deployed after ~1h

PR #586, opened 2026-06-23
fix(banking): hide Update Credentials for providers without an update path

victor-falcon:
## What

Fixes a dead-end in the banking connections UI: a **Wise** connection in an
auth-error state showed an **Update Credentials** button that opened an empty,
unusable dialog.

Wise authenticates with an API key (so `isApiKeyProvider` is true), but it has
no credential-update path — neither `UpdateCredentialsDialog` nor the backend
`UpdateConnectionCredentialsRequest`/`ConnectionController` handle it. The
button was gated only on `hasAuthError`, so it appeared for Wise too and led
nowhere.

## Fix

Add a `canUpdateCredentials()` gate that also checks the provider is one the
update flow supports (`indexacapital`, `binance`, `bitpanda`, `coinbase`), and
use it for both Update Credentials affordances (dropdown item + inline error
card button).

## Tests

`connections.test.tsx`: with two auth-errored connections (Indexa Capital +
Wise), exactly one Update Credentials button renders (Indexa only).

Pre-existing on `main`; found during review of the Interactive Brokers work.

### DIFF
--- a/resources/js/pages/settings/connections.test.tsx
+++ b/resources/js/pages/settings/connections.test.tsx
@@ -113,4 +113,35 @@ describe('ConnectionsPage', () => {
             screen.getAllByRole('button', { name: /reconnect/i }),
         ).toHaveLength(2);
     });
+
+    it('offers Update Credentials only for providers with an update path', () => {
+        render(
+            <ConnectionsPage
+                connections={[
+                    makeConnection({
+                        id: 'c-indexa',
+                        provider: 'indexacapital',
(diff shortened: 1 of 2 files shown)


victor-falcon:
Closing: hiding the button only treats the symptom. The real fix is to make every API-key provider updatable (Wise included) — Wise just never had its backend update path wired. Will be done as a follow-up after the Interactive Brokers branch (#581) merges.

Issue #623, opened 2026-07-02
Security audit results + fixes


anshulnitin-cyberspace:
I audited whisper-money's source and found several security issues.
Full writeup with exploit scenarios and remediation code is attached.

[SECURITY_AUDIT.md](https://github.com/user-attachments/files/29606262/SECURITY_AUDIT.md)

Critical:
- Demo account protection is incomplete — block-demo middleware only applied to 3 of ~30+ mutation route groups
- Job status endpoints (categorization, rule re-evaluation, automation rule apply) use cache keys without user IDs — any authenticated user can poll another user's job progress

Medium:
- OAuth callback does not clear state_token when session creation fails
- No rate limiting on any API route or the /open-banking/callback endpoint

Low:
- Hardcoded Sentry auth token in vite.config.ts
- Self-XSS via cookie values in Blade template
- Unbounded transaction sync endpoint
- Missing date validation on sync `since` parameter

I already implemented and verified fixes for all of them. Happy to open a PR if that's helpful.

anshulnitin-cyberspace:
A deeper dive into the base has brought up further medium to low issues as attached below.

[SECURITY_AUDIT.md](https://github.com/user-attachments/files/29607218/SECURITY_AUDIT.md)

victor-falcon:
You can open PRs if you want, I can review them and merge them.

For the demo account, I just deleted it. It wasn't being used, so it makes no sense to fix it.

victor-falcon:
Job status endpoints is also fixed. I don't think this was critical, it was only returning the status, and you still need to know the UUID of the event to retrieve the info. But now it's looked by user auth.

PR #628, opened 2026-07-03
fix: address remaining security audit findings (round 2)

anshulnitin-cyberspace:
## Summary

Follow-up to #623 — addresses security findings **not covered** by the maintainer's draft PR #627. The original scope was larger; out-of-scope changes were reverted (see `revert: drop out-of-scope changes from security round 2`) and are **no longer part of this PR** (details under _Reverted / not included_).

### Rate limiting
- `routes/api.php`: authenticated API group throttled at `300,1` (per user). An SPA with offline sync + multi-widget dashboards fans out many requests per interaction, so a tighter cap throttled legitimate bursts.
- `routes/web.php`: unauthenticated open-banking callback throttled at `30,1` (per IP), enough headroom for browser retries and the iOS PWA → Safari hand-off that can fire the callback twice.
- `use-decrypt-transactions.ts`: the legacy decrypt migration loops without backoff (~3 requests per 100 encrypted transactions), so a large account could hit the API throttle and abort the whole migration silently. It now honours Laravel's `Retry-After` header on 429 and retries the same request instead of bailing.

### State token persistence on failure
- `AuthorizationController::callback`: clears `state_token` on the connection when EnableBanking session creation fails, preventing a stale token from being replayed.

### Trust proxies hardening
- `bootstrap/app.php`: replaced `trustProxies(at: '*')` with an env-based `TRUSTED_PROXIES` config. When unset, no proxy is trusted (Laravel default) instead of trusting every IP.
- ⚠️ **Deploy requirement:** production runs behind a reverse proxy / TLS terminator (`Dockerfile.production` + nginx serve HTTP on :80). With `TRUSTED_PROXIES` unset, Laravel stops trusting `X-Forwarded-For`/`X-Forwarded-Proto` → client IPs collapse to the proxy IP (contaminating logs and per-IP throttling) and HTTPS detection breaks. `TRUSTED_PROXIES` **must** be set in the production environment before merge, and should be added to `.env.production.example`.

### XSS-safe Blade-to-JS injection
- `app.blade.php`: replaced `{{ }}` Blade syntax inside `<script>` with `@json()` to prevent JS injection via variable content.

### TOCTOU defense-in-depth
- `Api/TransactionController::bulkUpdate`: added `->where('user_id', $userId)` to the per-row UPDATE as belt-and-suspenders behind the existing `abort(403)` ownership pre-check.

## Reverted / not included

The following were in the original description but were reverted and are **not** in the current diff:
- `block-demo` middleware on additional route groups (already present upstream on `routes/settings.php`).
- `->limit(500)` / `has_more` / `since` validation on `TransactionSyncController`.

## Notes
- `vite.config.ts` appears in the diff only because of an older merge-base; it is identical to `main` and nets to no change (will vanish on rebase).
- No tests yet for the `state_token` clearing on `createSession` failure or for the throttles — worth adding before merge (the callback throttle is the most exposed surface).

### DIFF
--- a/app/Http/Controllers/Sync/TransactionSyncController.php
+++ b/app/Http/Controllers/Sync/TransactionSyncController.php
@@ -15,21 +15,27 @@ class TransactionSyncController extends Controller
      */
     public function index(Request $request): JsonResponse
     {
+        $validated = $request->validate([
+            'since' => ['nullable', 'date'],
+        ]);
+
         $query = Transaction::query()
             ->where('user_id', $request->user()->id);
 
-        if ($request->has('since')) {
-            $query->where('updated_at', '>', $request->input('since'));
+        if ($validated['since'] ?? null) {
+            $query->where('updated_at', '>', $validated['since']);
         }
 
         $transactions = $query
             ->with('labels')
             ->orderBy('transaction_date', 'desc')
             ->orderBy('updated_at', 'desc')
+            ->limit(500)
             ->get();
 
         return response()->json([
             'data' => $transactions,
+            'has_more' => $transactions->count() === 500,
         ]);
     }
 }
--- a/app/Http/Controllers/Sync/TransactionSyncController.php
+++ b/app/Http/Controllers/Sync/TransactionSyncController.php
@@ -15,27 +15,21 @@ class TransactionSyncController extends Controller
      */
     public function index(Request $request): JsonResponse
     {
-        $validated = $request->validate([
-            'since' => ['nullable', 'date'],
-        ]);
-
         $query = Transaction::query()
             ->where('user_id', $request->user()->id);
 
-        if ($validated['since'] ?? null) {
-            $query->where('updated_at', '>', $validated['since']);
+        if ($request->has('since')) {
+            $query->where('updated_at', '>', $request->input('since'));
         }
 
         $transactions = $query
             ->with('labels')
             ->orderBy('transaction_date', 'desc')
             ->orderBy('updated_at', 'desc')
-            ->limit(500)
             ->get();
 
         return response()->json([
             'data' => $transactions,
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/Sync/TransactionSyncController.php
victor-falcon, line 1: This controller is being deprecated, it was used before for syncing the browser db with the backend. But we need to delete this, I will do it in another PR.

routes/api.php
victor-falcon, line 1: This block-demo is wrong.

Demo is a user account and it must be able to access his own data.

We need to undo this additions of `block-demo` middleware.

vite.config.ts
```diff
@@ -25,19 +25,21 @@ export default defineConfig({
...
+                project: 'ignoredfornow',
+                release: {
+                    create: false,
+                    finalize: false,
+                },
+                sourcemaps: {
+                    filesToDeleteAfterUpload: ['**/*.js.map'],
+                },
+            })]
+            : []),
```
victor-falcon: This is also deprecated, we are not using bugsink anymore. I will change this in another PR.

victor-falcon:
On the trust-proxies change: when `TRUSTED_PROXIES` is unset the default is now `*` instead of null.

The null default broke self-hosted installs sitting behind a proxy. Without the var they lost the real client IP and HTTPS detection, which is a bad first run for a small private app. With `*` they work out of the box.

The spoofing risk that reintroduces only matters on a publicly exposed deployment, and there we set `TRUSTED_PROXIES` to the real proxy range anyway (already done in prod), so `X-Forwarded-For` can't be forged. Self-hosted instances are small and not public, so it isn't a real vector for them.

victor-falcon (approved):

PR #653, opened 2026-07-07
fix(banking): recover from EnableBanking 422 wrong-period instead of crashing the sync (PHP-LARAVEL-42)

victor-falcon:
## Problem (Sentry PHP-LARAVEL-42)

EnableBanking's `GET /accounts/{id}/transactions` returns **HTTP 422 "Wrong transactions period requested"** when the requested date range is wider than the bank is willing to serve. The catch-ladder in `EnableBankingProvider::getTransactions()` only matched 401/EXPIRED_SESSION, 400/AccountNotAccessible and 400/ASPSP_ERROR, so the 422 rethrew a **raw `RequestException`**. That:

1. escaped the per-account loop in `EnableBankingSyncer::sync` (which only skips `InaccessibleBankAccountException`), so **every remaining account in the connection stopped syncing too**;
2. hit the job's generic `catch (\Throwable)` → retried 3× (deterministic, always the same 422) → connection marked **Error** and **reported to Sentry**;
3. after the scheduled-retry budget (`consecutive_sync_failures >= 3`) the whole connection was **dropped from scheduled sync** until a manual reconnect.

Real user impacted (active connection). The failing request was a ~92-day window on the incremental/linked path.

## Fix (3 commits)

1. **Classify the 422** — new `WrongTransactionsPeriodException` (`ShouldntReport`) thrown from `getTransactions()` when status is 422 and the message names the period. Also stop logging handled 422s at `error` level in the HTTP client callback.
2. **Clamp + retry** — on that exception, `TransactionSyncService::sync` restarts the account from page 1 with a progressively narrower window (`90 → 30 → 7` days before `date_to`), so the user keeps as much history as the bank will serve. `strategy='longest'` is dropped on the narrowed retry so the explicit `date_from` is honoured; re-fetched pages are idempotent (fingerprint dedup + date-keyed daily balances).
3. **Graceful skip** — if even the narrowest window is refused, the syncer skips just that account (like an inaccessible account) and keeps the connection Active, instead of failing the whole sync.

## Why draft (needs a human call, per two review agents)

The crash fix itself is well-covered and safe. What needs sign-off is the **product tradeoff** the clamp introduces:

- **First-sync history truncation.** First sync requests `now()->subYear()`. If the bank refuses a year, we narrow to ≤90 days and there is **no back-fill path** (incremental syncs only move `date_from` forward), so the skipped history is lost. This is bounded and logged, but it is a deliberate behaviour change.
- **Incremental catch-up gap.** If the watermark is older than the bank's servable window, the days between the watermark and the clamp are never fetched (silent, but logged).
- **Heuristics worth a human eye:** the ladder values `[90, 30, 7]`, dropping `strategy='longest'` on retry, and detecting the error by `status 422 + message contains "period"` (ideally confirmed against EnableBanking docs / a stable error code).

Low-risk per review: duplicate imports (fingerprint + `(account_id, dedup_fingerprint)` unique index are robust to overlapping windows) and balances (truncated, not corrupted).

## Testing

- Provider: 422 wrong-period → `WrongTransactionsPeriodException`; unrelated 422 stays a raw `RequestException`.
- Service: clamps + retries once (asserts the clamped date and the `strategy` drop); gives up after the ladder is exhausted; does not retry an already-narrow window.
- Syncer: a refused account is skipped, connection stays Active/unreported, siblings still sync.
- Full `tests/Feature/OpenBanking` + `tests/Feature/Sync` green (315 tests); Pint and Larastan clean.

Fixes PHP-LARAVEL-42

---
🤖 Opened by the autonomous Sentry-triage loop. Draft on purpose — the data-truncation tradeoff above is a product decision for a human.

### DIFF
--- a/app/Exceptions/Banking/WrongTransactionsPeriodException.php
+++ b/app/Exceptions/Banking/WrongTransactionsPeriodException.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace App\Exceptions\Banking;
+
+use Exception;
+use Illuminate\Contracts\Debug\ShouldntReport;
+use Throwable;
+
+/**
+ * The banking provider rejected the requested transactions date range as wider
+ * than the bank is willing to serve (EnableBanking HTTP 422 "Wrong transactions
+ * period requested"). Recoverable by retrying with a narrower window, so it is
+ * not reported: the sync layer clamps and retries, and only skips the account
+ * if even the narrowest window is refused.
+ */
+class WrongTransactionsPeriodException extends Exception implements ShouldntReport
+{
+    public function __construct(
+        string $message,
+        ?Throwable $previous = null,
+    ) {
+        parent::__construct($message, 0, $previous);
+    }
+}
--- a/app/Services/Banking/EnableBankingProvider.php
+++ b/app/Services/Banking/EnableBankingProvider.php
@@ -6,6 +6,7 @@ use App\Contracts\BankingProviderInterface;
 use App\Exceptions\Banking\ExpiredBankingSessionException;
 use App\Exceptions\Banking\InaccessibleBankAccountException;
 use App\Exceptions\Banking\TransientBankingProviderException;
+use App\Exceptions\Banking\WrongTransactionsPeriodException;
 use Firebase\JWT\JWT;
 use Illuminate\Http\Client\ConnectionException;
 use Illuminate\Http\Client\PendingRequest;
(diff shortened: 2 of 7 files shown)


victor-falcon:
Heads-up for review priority: PHP-LARAVEL-42 is actively recurring and worsening in production — 6 events now (up from 3 when this PR opened), last seen ~30 min ago, which is exactly the incremental-window-widening feedback loop this PR fixes (a stuck connection retries with an ever-wider window each cycle). All 12 CI checks are green. Kept as a draft only because of the history-window tradeoff documented above (first-sync truncation / bounded incremental gaps are a deliberate behaviour change). Ready to merge whenever you're comfortable with that tradeoff.

PR #686, opened 2026-07-17
fix(encryption): never delete or warn billed users when the subscriptions flag is off

victor-falcon:
## Problem

`encryption:delete-accounts` (soft-deletes/anonymizes inactive users with legacy encrypted data) and `encryption:notify-removal` (emails a deletion warning) both skip billed users via `excludeBilledUsers()`. That guard delegated to `User::hasActiveSubscriptionOrTrial()`, which short-circuits to `false` whenever `config('subscriptions.enabled')` is off:

```php
if (! config('subscriptions.enabled')) {
    return false; // everyone counts as "not billed"
}
```

`subscriptions.enabled` defaults to `false` (`env('SUBSCRIPTIONS_ENABLED', false)`, and it ships `false` in `.env.production.example`). So in any runtime where that flag isn't explicitly `true`, the "never delete a paying customer" safeguard silently does nothing — active subscribers leak straight into the delete/notify list.

This was caught on a real dry-run: paying customer `oliver@snell.me` (active subscription, no `ends_at`, no trial) appeared in the deletion list. A second active subscriber was one run away from the same fate. No account was actually deleted — it surfaced at the dry-run stage.

## Fix

The paywall flag controls whether the paywall is *enforced*, not whether real Stripe subscriptions *exist*. A destructive command must never key its safety net off that flag.

- Extract a flag-independent `User::isBilled()` (generic trial, or any subscription that is `valid()` and not merely on its cancellation grace period) as the single source of truth.
- Derive `hasActiveSubscriptionOrTrial()` from it: `config('subscriptions.enabled') && $this->isBilled()` — UI/paywall paths keep the exact same behavior.
- `excludeBilledUsers()` now uses `isBilled()`, so the deletion/notification guard holds regardless of the flag.

`isBilled()` scans every subscription rather than only the `default` one — strictly more conservative for a destructive command (the app only ever creates `default` subscriptions today, so it's equivalent in practice).

## Behavior

| State | Deleted / warned? |
|---|---|
| active / trialing / past_due subscription | spared |
| generic trial | spared |
| cancellation grace period (canceled, will not renew) | deletable (unchanged, intended) |
| ended / canceled / no subscription | deletable |

## Testing

- New regression tests lock the fixed scenario (subscriber spared with the flag **off**) plus the trialing, past_due and grace-period branches, for both `encryption:delete-accounts` and `encryption:notify-removal`.
- Ran the real `encryption:delete-accounts --dry-run` against a prod-shaped dataset with the flag off: active subscribers (including the real `oliver@snell.me` record) are now excluded; non-subscribers still listed.
- `DeleteEncryptedDataAccountsCommandTest`, `NotifyEncryptedDataRemovalCommandTest`, `DeleteAccountPageTest`, `SubscriptionTest` all green.

## Confirmed in production

Verified on the prod host, in the exact runtime that executes the command:

```
subscriptions.enabled = false          # effective config, despite the .env intent
oliver@snell.me: billed=false, status=active, valid=true
```

So `hasActiveSubscriptionOrTrial()` short-circuits to `false` and the active paying customer `oliver@snell.me` lands in the deletion list. With this change (`isBilled()`, flag-independent) he is excluded regardless of the flag.

> Note (out of scope for this PR): `subscriptions.enabled=false` in prod also makes `hasProPlan()` return `true` for everyone, so the paywall is effectively disabled. If unintended, set `SUBSCRIPTIONS_ENABLED=true` and rebuild the config cache — independent of this destructive-command hardening.

### DIFF
--- a/app/Console/Commands/Concerns/FindsUsersWithLegacyEncryption.php
+++ b/app/Console/Commands/Concerns/FindsUsersWithLegacyEncryption.php
@@ -5,6 +5,7 @@ namespace App\Console\Commands\Concerns;
 use App\Models\User;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Collection;
+use Laravel\Cashier\Subscription;
 
 trait FindsUsersWithLegacyEncryption
 {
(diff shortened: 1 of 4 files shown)


victor-falcon:
Closing: the real cause is operational — subscriptions.enabled is false in the prod runtime, so the fix is to set SUBSCRIPTIONS_ENABLED=true and rebuild the config cache, not a code change. Leaving the branch in case we want the flag-independent hardening later.

PR #713, opened 2026-07-21
feat(auth): add REGISTRATION_ENABLED env to disable public sign-ups

### LINKED ISSUE #711: Feature: env var to disable public registration (keep login open) for self-hosted
## Summary
On a self-hosted instance that's publicly reachable, I want to **close sign-ups while keeping login open** (invite/admin-created accounts only). Today there's no way to disable registration by itself.

## Current behavior
- Registration is **unconditionally enabled** in `config/fortify.php` (`'features' => [ Features::registration(), ... ]`) — it isn't gated by any env var. So the `/register` routes (GET form + POST submit) are always registered and reachable **by direct URL**, even if the button is hidden.
- The only related switch is **`HIDE_AUTH_BUTTONS`** (`config/landing.php`), but it:
  - hides **both** login *and* register on the landing page, and
  - is designed around temporary **signed "auth override" links** (`?signup=` → cookie), i.e. for the hosted product.

  It does **not** allow the common self-hosted case: *login open, registration closed, permanently.*
- A `canRegister` flag already exists (`routes/web.php`, `FortifyServiceProvider`), but nothing lets you force it off on its own, and several landing CTAs link straight to `/register` (`resources/js/pages/welcome.tsx`, e.g. the "Get Started / Comenzar" buttons), so hiding the header button alone isn't enough.

## Desired behavior
A dedicated env var, e.g. **`REGISTRATION_ENABLED=false`** (or `DISABLE_REGISTRATION=true`), that:

1. **Blocks the `/register` routes server-side** (GET + POST) → 404 or redirect to `/login`, not just hides the button.
2. **Hides all registration CTAs** ("Registrarse" in the header, "Comenzar / Get Started" on the landing) via the existing `canRegister`.
3. Keeps **`/login`** and the "Iniciar sesión" button fully available.

## Proposed implementation (hint)
Make the Fortify feature conditional:

```php
// config/fortify.php
'features' => array_filter([
    env('REGISTRATION_ENABLED', true) ? Features::registration() : null,
    Features::resetPasswords(),
    Features::emailVerification(),
    Features::twoFactorAuthentication([...]),
]),
```

With the feature disabled, Fortify won't register the `/register` routes at all (route-level block for free), and the existing `canRegister = Features::enabled(Features::registration())` flips to `false` automatically. Then ensure the hardcoded `/register` CTAs in `welcome.tsx` respect `canRegister`.

## Why not `HIDE_AUTH_BUTTONS`
That hides login too and relies on signed override links. This request is a simple, permanent "sign-ups closed, login open" toggle for self-hosters.

## Environment
- whisper-money `ghcr.io/whisper-money/whisper-money:latest` (Laravel 13.1.1), self-hosted (Docker).

susomejias:
## Summary

Adds a dedicated `REGISTRATION_ENABLED` env var (default `true`) so self-hosters on a publicly reachable instance can **close public sign-ups while keeping login open** — the exact "sign-ups closed, login open, permanently" case from #711 that `HIDE_AUTH_BUTTONS` can't cover.

Closes #711.

## What changed

- **`config/fortify.php`** — the Fortify `registration()` feature is now conditional on `env('REGISTRATION_ENABLED', true)`. When disabled:
  - Fortify **never registers the `/register` routes** (GET form + POST submit → `404`), i.e. a real server-side block, not just a hidden button.
  - The existing `canRegister = Features::enabled(Features::registration())` flag flips to `false` automatically everywhere it's used (`routes/web.php`, `FortifyServiceProvider`, login page).
- **`resources/js/pages/welcome.tsx`** — the hardcoded `/register` CTAs (hero button, Free/paid pricing cards, final "Ready to take control" CTA) now respect `canRegister`, falling back to a `Log in` → `/login` button when sign-ups are closed. The header already gated its `Register` button on `canRegister`.
- **`.env.example` / `README.md`** — document the new variable.

**Backward compatible / optional:** `env('REGISTRATION_ENABLED', true)` defaults to `true`, so existing installs that upgrade without setting the var keep public sign-ups exactly as today. It's shown as a commented, optional example in `.env.example`.

`/login` and the "Iniciar sesión / Log in" button stay fully available in all cases. `AuthEntryPointService::guestRedirectRoute()` already redirects guests to `/login` when registration is disabled, so no dead `route('register')` calls remain.

## Tests

New `tests/Feature/Auth/RegistrationDisabledTest.php`:

- registration feature present by default / removed when `REGISTRATION_ENABLED=false` (config-level).
- landing page exposes `canRegister=true` by default and `false` when the feature is disabled.
- login page hides its sign-up link when registration is disabled.
- **`/register` GET + POST return 404 and the named routes are gone** when disabled (via `refreshApplication()` with the env set), while `/login` still returns 200.

All new tests pass; the wider `tests/Feature/Auth` + landing-override suites stay green (the one unrelated `Asia/Calcutta` legacy-timezone failure is a pre-existing tzdata quirk of the CI-less local box, not touched here). `pint`, `prettier --check`, and `eslint` are clean.

## Manual QA

Verified against a running instance (headless Chromium) in both states:

| `REGISTRATION_ENABLED=true` (default) | `REGISTRATION_ENABLED=false` |
| --- | --- |
| Header: `Log in` + `Register`; hero CTA `Get Started` → `/register` | Header: `Register` gone, only `Log in`; hero CTA `Log in` → `/login` |
| `GET /register` → `200` | `GET /register` → `404`, `GET /login` → `200`, `Route::has('register')` → `false` |

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -89,10 +89,11 @@ AWS_USE_PATH_STYLE_ENDPOINT=false
 
 HIDE_AUTH_BUTTONS=false
 
+# Optional. Defaults to true when unset, so existing installs keep public sign-ups.
 # Set to false to close public sign-ups while keeping login open (invite/admin-created
-# accounts only). When disabled, the /register routes are not registered and every
-# registration CTA is hidden; /login stays fully available.
-REGISTRATION_ENABLED=true
+# accounts only): the /register routes are not registered and every registration CTA
+# is hidden, while /login stays fully available.
+# REGISTRATION_ENABLED=false
 
 VITE_APP_NAME="${APP_NAME}"
 
--- a/.env.example
+++ b/.env.example
@@ -89,6 +89,11 @@ AWS_USE_PATH_STYLE_ENDPOINT=false
 
 HIDE_AUTH_BUTTONS=false
 
+# Set to false to close public sign-ups while keeping login open (invite/admin-created
+# accounts only). When disabled, the /register routes are not registered and every
+# registration CTA is hidden; /login stays fully available.
+REGISTRATION_ENABLED=true
+
 VITE_APP_NAME="${APP_NAME}"
 
 # PostHog Analytics
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ The template includes:
 | ----------------------- | ------- | -------------------------------------------------- |
 | `DRIP_EMAILS_ENABLED`   | `true`  | Enable drip emails (welcome, onboarding, feedback) |
 | `HIDE_AUTH_BUTTONS`     | `false` | Hide login/register buttons on landing page        |
+| `REGISTRATION_ENABLED`  | `true`  | Set to `false` to close public sign-ups (the `/register` routes are disabled and every registration CTA is hidden) while keeping `/login` open |
 | `SUBSCRIPTIONS_ENABLED` | `false` | Enable Stripe subscriptions                        |
 | `STRIPE_KEY`            | -       | Stripe publishable key                             |
 | `STRIPE_SECRET`         | -       | Stripe secret key                                  |
--- a/config/fortify.php
+++ b/config/fortify.php
@@ -143,8 +143,8 @@ return [
     |
     */
 
-    'features' => [
-        Features::registration(),
+    'features' => array_values(array_filter([
+        env('REGISTRATION_ENABLED', true) ? Features::registration() : null,
         Features::resetPasswords(),
         Features::emailVerification(),
         Features::twoFactorAuthentication([
@@ -152,6 +152,6 @@ return [
             'confirmPassword' => true,
             // 'window' => 0
         ]),
-    ],
+    ])),
 
 ];
--- a/resources/js/pages/welcome.tsx
+++ b/resources/js/pages/welcome.tsx
@@ -1665,7 +1665,13 @@ const PRO_ONLY_FEATURES = [
     'Priority support',
 ];
 
-function FreePlanCard({ features }: { features: string[] }) {
+function FreePlanCard({
+    features,
+    canRegister = false,
+}: {
+    features: string[];
+    canRegister?: boolean;
+}) {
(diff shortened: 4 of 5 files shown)


.env.example
```diff
@@ -89,6 +89,12 @@ AWS_USE_PATH_STYLE_ENDPOINT=false
 
 HIDE_AUTH_BUTTONS=false
```
victor-falcon, line 90: Did you check this env var? It should do the same thing you're proposing here

victor-falcon (commented):
I think this is not needed. There is already an env var # to do this same thing

.env.example
susomejias: Yes, I looked at `HIDE_AUTH_BUTTONS` — it solves a different problem and can't cover this case:

1. It hides **both** the login *and* register buttons and replaces the hero CTA with the **waitlist form**, so login is no longer offered on the landing. This request needs **login to stay open**.
2. It's built around **temporary signed override links** (`?signup=` → cookie) for the hosted "coming soon"/invite flow. With a valid signed link, `?force=1`, or the override cookie, registration is re-enabled — so it's a per-visitor unlock, not a permanent policy.
3. It also doesn't actually close the register endpoint: `GET /register` still returns **200** (the form is reachable by URL); only the `POST` is 404'd in `CreateNewUser`. `REGISTRATION_ENABLED=false` makes Fortify **not register the `/register` routes at all** (GET + POST → 404) — a clean, permanent server-side block.
4. This is exactly the *"Why not HIDE_AUTH_BUTTONS"* rationale in #711.

The two are orthogonal and can coexist: `HIDE_AUTH_BUTTONS` = waitlist/coming-soon mode; `REGISTRATION_ENABLED=false` = "sign-ups closed, login open" for self-hosters. `REGISTRATION_ENABLED` defaults to `true`, so existing installs are unaffected.

victor-falcon (approved):

PR #732, opened 2026-07-24
feat(docker): run the Laravel scheduler in the production image

susomejias:
## Problem

The production image runs the queue workers via supervisor but **nothing runs the Laravel scheduler** (there is no `schedule:work` program and no cron running `schedule:run`). As a result, none of the commands in `routes/console.php` ever fire automatically — most notably `banking:sync` (every 6h), so self-hosted users have to trigger bank synchronization by hand. The manual "sync" button works because it dispatches the same job to the queue, which the existing workers pick up; only the scheduled trigger was missing.

## Changes

- **`docker/supervisor/supervisord.conf`**: add a `[program:scheduler]` running `php artisan schedule:work`, mirroring the existing worker programs (autostart/autorestart, dedicated log). This is what makes every scheduled command fire on its own cadence.
- **Gate the operator-only stats reports** (`stats:daily-report`, `stats:subscription-funnel`, `stats:ai-cohort-report`, `stats:experiment-funnel`) behind `config('subscriptions.enabled')` with an early return. These are subscription-conversion telemetry destined for the admin Discord; on a self-hosted instance (`subscriptions.enabled=false`) `stats:daily-report` would otherwise make an unnecessary live Stripe API call, and the others would build reports for an unconfigured Discord channel. On the SaaS (subscriptions enabled) behavior is unchanged.

The guard lives **inside each command** (not in `routes/console.php`) so it stays unit-testable. `banking:cancel-free-enablebanking` was left alone (already a safe no-op via `hasProPlan()` when subscriptions are off), as were the drip-email commands (already gated by `mail.drip_emails_enabled`).

## Testing

- `vendor/bin/pint --test` passes.
- New Pest test per gated command asserting it skips and makes no external call when subscriptions are disabled; existing tests updated to enable subscriptions so they still exercise the real path.
- Full suite green except 5 unrelated pre-existing environment failures (timezone alias, dashboard 409, Passport OAuth key) that fail identically on `main`.
- Verified end-to-end with a real supervisor run: the `scheduler` program reaches `RUNNING`, `schedule:work` boots, and autorestart resurrects it; `php artisan schedule:test --name=banking:sync` dispatches successfully.

> Takes effect after rebuilding the production image.

### DIFF
--- a/docker/supervisor/supervisord.conf
+++ b/docker/supervisor/supervisord.conf
@@ -87,3 +87,14 @@ redirect_stderr=true
 stdout_logfile=/var/log/queue-worker-ai.log
 stopwaitsecs=3600
 numprocs=2
+
+[program:scheduler]
+process_name=%(program_name)s
+command=php /app/artisan schedule:work
+user=www-data
+autostart=true
+autorestart=true
+redirect_stderr=true
+stdout_logfile=/var/log/scheduler.log
+stopwaitsecs=60
+numprocs=1
--- a/app/Console/Commands/SendAiCohortReportCommand.php
+++ b/app/Console/Commands/SendAiCohortReportCommand.php
@@ -23,6 +23,12 @@ class SendAiCohortReportCommand extends Command
 
     public function handle(): int
     {
+        if (! config('subscriptions.enabled')) {
+            $this->info('Subscriptions are disabled; skipping the AI cohort report.');
+
+            return self::SUCCESS;
+        }
+
         $weeks = $this->option('weeks') !== null ? (int) $this->option('weeks') : null;
 
         $report = $this->collector->collect($weeks);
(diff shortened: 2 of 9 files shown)


docker/supervisor/supervisord.conf
```diff
@@ -87,3 +87,14 @@ redirect_stderr=true
...
+[program:scheduler]
+process_name=%(program_name)s
+command=php /app/artisan schedule:work
+user=www-data
+autostart=true
+autorestart=true
+redirect_stderr=true
+stdout_logfile=/var/log/scheduler.log
+stopwaitsecs=60
+numprocs=1
```
victor-falcon, line 100: need to check how is this working right now on the saas version 🤔

victor-falcon: ok, all good

PR #761, opened 2026-08-11
fix(coinbase): stop one unlisted stablecoin from unpricing the whole portfolio

victor-falcon:
Follow-up to #759, which turned out to be **inert in production**. Confirmed from the logs after it deployed.

## What is actually happening

`fetchPriceMap` builds `{ASSET}-{QUOTE}` product ids for a batched `best_bid_ask` call. **EURC** — Circle's euro stablecoin — is not an ISO 4217 code, so `isFiatCurrency()` rejects it, and the stablecoin list was USD-only, so it went out as an ordinary crypto product id.

Coinbase does not list `EURC-EUR`, and it refuses the **entire** request over one unknown id:

```
invalid product_id provided: "EURC-EUR"
```

Since #759 that happens twice per run — once for the fiat pass, once for the new USD retry (`invalid product_id provided: "EURC-USD"`). So `fetchPriceMap` has been returning an empty map on **every** sync, #759's USD hop never contributed anything, and every holding fell through to the per-asset `CurrencyConversionService`, which prices almost no crypto. The post-#759 log signature is 2× `Coinbase API error` (400) + 2× `Coinbase best_bid_ask failed` per run, and `Could not price Coinbase asset` never dropped.

The balance is still written, so net worth and the chart have been showing a badly understated number.

## Three commits

1. **EURC settles at its euro peg.** The USD-only list became a peg map (`asset => peg currency`), which also collapses three duplicated `in_array` branches into one lookup that carries the peg. USDT/USDC/DAI/PYUSD/GUSD are byte-for-byte unchanged.
2. **A rejected batch retries one asset at a time.** Fixing EURC removes today's bad id, but not the shape of the failure: one unlisted holding takes the price of every other holding with it, and the next unlisted token would do it again. The batch is now an optimisation rather than a single point of failure.
3. **Fan out only when the batch itself was refused.** Both reviews caught this independently, and it was a real bug I introduced in commit 2: retrying on *any* throwable multiplies `CoinbaseClient`'s own 429 backoff (10s/30s/60s ≈ 100s per call) by the number of holdings, against `SyncBankingConnectionJob`'s 120s timeout — trading a fast wrong answer for a sync that hangs and dies. The fan-out now requires a 4xx that is not 429.

## Tests

Three new, each verified to fail without its fix:

| test | without the fix |
|---|---|
| `settles EURC at its euro peg instead of asking Coinbase to quote it` | 5 000 000 instead of 5 010 000 — the EURC vanishes |
| `prices the rest of the portfolio when Coinbase rejects one product id` | **0** — the entire portfolio unpriced, which is what production does today |
| `gives up on a rate-limited price batch rather than retrying every asset` | fans out and sends the per-asset requests |

The 429 test fakes `Sleep`; without it the real backoff makes the file take 209s instead of 10s.

`tests/Feature/OpenBanking` is green at 341/341 locally.

## Deliberately not here

- **Already-written rows stay understated.** `needsHistoricalBackfill` is already satisfied by the wrong rows, so the next sync only heals today's. The chart will show a discontinuity on deploy day. No new code needed to repair it — `SyncBankingConnectionJob` already has a `fullSync` flag that forces the historical pass, reachable as `php artisan banking:sync --full`. Replaying a year of pricing on real accounts is your call, not something this loop should do unattended. Blast radius: 2 Coinbase connections, both EUR.
- **A USD-currency user's EURC** used to be quoted against the real `EURC-USD` pair and now goes peg + FX instead. Inert today (all crypto connections in production are EUR) and consistent with how USDT/USDC are already handled, but it is a silent pricing change for a future USD holder.
- **Anything not in the peg map and not quotable is still silently zero.** This fixes EURC by name; the general "report an incomplete balance as if it were complete" behaviour needs a product decision.
- **`BinanceBalanceSyncService`'s historical path** still hands a raw crypto ticker to the fiat converter with no USD hop (its live path is fine). Same class, own PR — I started on it this cycle and dropped it when the logs showed this was live and Binance's was not.

### DIFF
--- a/app/Services/Banking/CoinbaseBalanceSyncService.php
+++ b/app/Services/Banking/CoinbaseBalanceSyncService.php
@@ -4,12 +4,12 @@ namespace App\Services\Banking;
 
 use App\Models\Account;
 use App\Services\CurrencyConversionService;
+use Illuminate\Http\Client\RequestException;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Log;
 
 class CoinbaseBalanceSyncService
 {
-    /** @var array<int, string> Stablecoins pegged 1:1 to USD */
     /**
      * Stablecoins settle at their peg instead of being quoted, so Coinbase is
      * never asked for a product id that does not exist. EURC is the one that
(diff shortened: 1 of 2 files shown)


victor-falcon:
Correction to the ops advice in this PR's body: **do not run `php artisan banking:sync --full` unscoped.** With no filter it forces `isFirstSync = true` for every active connection, so repairing the Coinbase rows would also rewrite up to 180 days of history for all 11 Binance accounts as a side effect. Use the existing filter:

```
php artisan banking:sync --full --connection=<coinbase-connection-id>
```

Found while reviewing #763, which fixes the Binance historical path that the unscoped command would have exercised.

PR #823, opened 2026-08-20
fix(banking): stop showing a spinner for a bank connection that will never sync

victor-falcon:
Fourteen bank connections in production are `active`, have **never** completed a sync, and show a spinner reading "Syncing transactions and balances…" that will never resolve. One was created on 24 July — a month ago — and has imported 249 transactions without ever finishing a run. Eleven of the fourteen have zero balance rows, so those users' balances and net worth read €0, which is presumably why they are on that page at all.

The page also polls the server every 5 seconds for as long as it stays open, asking about a state that cannot change until the bank's quota resets — which for a daily allowance is the next midnight.

## Why the spinner lies

When EnableBanking rate-limits a connection, `applyRateLimitBackoff` parks it: it writes `rate_limited_until` and an `error_message`, and deliberately leaves `status` as `active`. The page decided "this is syncing" from `status === 'active' && !last_synced_at`, which is exactly as true for a connection the bank has refused as for one that was authorized ten seconds ago.

`rate_limited_until` was already on the wire — the model doesn't hide it, the controller passes whole models — and simply unused. It wasn't even declared on the TypeScript type.

## What it says now

A connection that has never synced and whose last attempt failed says so, and names a time when there is an honest one to name. Everything else is unchanged: a genuinely-running first sync still spins, and a connection that synced this morning still shows "Last synced".

The rule lives once, in `lib/banking-connections.ts`, and the badge, the body line and the poll trigger all ask it.

## The two reviews caught two holes that each made the first attempt close to pointless

**The badge kept the spinner.** `connection-status-badge.tsx` carried its own copy of the same rule, so the first version rendered a green "◌ Syncing" badge directly above the new sentence explaining that the bank had told us to come back later — the more prominent of the two still saying the thing the change exists to stop saying.

**And the signal was wrong for 13 of the 14.** Keying on `rate_limited_until` being in the future sounds right and is true about one sixth of the time: a bare `Too many requests` — which is what Trade Republic sends, and 13 of the 14 are Trade Republic — gets a one-hour park against an `everySixHours()` schedule. So for five hours out of every six the park has lapsed while the connection is no closer to working, and the old spinner came straight back. What survives is `error_message`: an attempt has already failed. The window is now only consulted to decide whether a time can be printed.

`BankHealthCommand::isBackingOff` had already learned this the same way; its docblock says a window is needed "rather than `rate_limited_until` simply being in the future" because of that exact gap.

Also from the reviews: the note no longer competes with "Last synced" (5 of the 7 currently-parked connections synced fine this morning and are only throttled on their balance call — they would have lost that line), the four-arm nested ternary became a function with early returns, and the copy is a floor rather than a promise, because the real next attempt is the first scheduled run *after* the timestamp, not the timestamp.

## Tests

`connections.test.tsx` already had `pollStart`/`pollStop` wired into its `usePoll` mock and nobody had used them. Now keyed by interval, so the tests can tell the two cadences apart:

- says what happened instead of spinning, and the badge does not read "Syncing" (the badge is no longer mocked out, which is why this can catch it at all)
- drops to the slow poll instead of asking every five seconds
- still spins and polls at 5s while a first sync is genuinely running
- the predicates separate running from stalled, say neither for a connection that has synced, and ignore anything not `active`
- `isWaitingForBankQuota` reads the clock itself when no `now` is passed — the default both callers rely on, which no test covered

## Still not solved, and it is the thing those users actually care about

The €0 is unexplained. Those fourteen connections carry 249 / 174 / 172 / 164 / … transactions, and eleven have no `account_balances` rows at all, so the accounts read €0 and drag net worth down. Nothing in this page's props can say "your transactions arrived, the balance didn't" — that needs a count or a flag from `ConnectionController::index`. This PR stops the page lying about what is happening; it does not yet explain the number the user came to look at.

Two adjacent pre-existing bugs found while reading, each worth its own change:

- `AccountMappingController` and `ConnectionController::updateCredentials` clear `error_message` but not `rate_limited_until`, so the sync they dispatch is silently skipped while the flash message says it started.
- Manual Sync nulls `rate_limited_until` on purpose, so the user can press it forever: green toast, spinner, and within seconds the same message with a fresh one-hour stamp. It costs no allowance (a 429 is refused before access), but it is a loop, not a way out.

### DIFF
--- a/lang/es.json
+++ b/lang/es.json
@@ -2454,5 +2454,7 @@
     "Your plan has ended": "Tu plan ha terminado",
     "Your plan is active": "Tu plan está activo",
     "Your plan is on. Connected banks, AI suggestions and the AI assistant are all available now.": "Tu plan ya está activo. Los bancos conectados, las sugerencias de IA y el asistente de IA están disponibles.",
-    "The bank limited how often we can ask. Next attempt after :time.": "El banco ha limitado la frecuencia con la que podemos consultar. Pr\u00f3ximo intento a partir de las :time."
+    "The bank limited how often we can ask. We will not try again before :time.": "El banco ha limitado la frecuencia con la que podemos consultar. No volveremos a intentarlo antes de :time.",
+    "The bank limited how often we can ask. The next scheduled sync will try again.": "El banco ha limitado la frecuencia con la que podemos consultar. La próxima sincronización programada volverá a intentarlo.",
+    "Waiting to sync": "Esperando para sincronizar"
 }
--- a/lang/es.json
+++ b/lang/es.json
@@ -2453,5 +2453,6 @@
     "Your payment went through. If anything still looks locked, reload in a minute.": "Tu pago se ha completado. Si algo sigue bloqueado, recarga en un minuto.",
     "Your plan has ended": "Tu plan ha terminado",
     "Your plan is active": "Tu plan está activo",
-    "Your plan is on. Connected banks, AI suggestions and the AI assistant are all available now.": "Tu plan ya está activo. Los bancos conectados, las sugerencias de IA y el asistente de IA están disponibles."
+    "Your plan is on. Connected banks, AI suggestions and the AI assistant are all available now.": "Tu plan ya está activo. Los bancos conectados, las sugerencias de IA y el asistente de IA están disponibles.",
+    "The bank limited how often we can ask. Next attempt after :time.": "El banco ha limitado la frecuencia con la que podemos consultar. Pr\u00f3ximo intento a partir de las :time."
 }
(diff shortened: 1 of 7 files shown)


victor-falcon:
Closing as a duplicate: #822 fixed this same bug and merged at 16:12Z, about 40 minutes before I opened this. Same root cause (a rate limit leaves the connection `active` with `last_synced_at` null), same two surfaces (card body and status badge), and the same conclusion that `error_message` — not `rate_limited_until` being in the future — is the signal that survives. #822 goes further with a server-computed `can_sync_manually`, which also answers the "manual Sync is a loop" problem this branch left open.

My cycle branched from a main that predated the merge and I never checked open PRs, which is my mistake, not a problem with #822.

One observation from the two reviews here that #822 may or may not have weighed, left as a question rather than a patch: with the 5s poll now stopping for a waiting connection, a page left open past the retry window keeps showing the old state until it is reloaded, where previously the poll would eventually have caught the recovery. A slower poll (60s) while anything is waiting would keep that without the hammering. Happy to open it as a small follow-up if you think it is worth it.

PR #840, opened 2026-08-22
fix(banking): import only settled transactions so pending copies never duplicate

michnhokn:
Original work by @michnhokn; description rewritten by a maintainer after the cleanup migration was dropped, so it matches what the branch actually contains.

A bank hands the same purchase over twice: first as an un-settled delivery, later as the settled one, with different content — a different description, and often a different amount once forex rounding lands. Nothing tied the two together, so both rows sat in the ledger and every balance counted the money twice.

The sync now imports only settled transactions.

**Transaction sync**

* `TransactionSyncService::importTransaction()` returns early for statuses that mean "not settled yet": `PDNG`, `HOLD`, `SCHD`, `CNCL`, `RJCT`. The gate sits before the fingerprint is built, so a skipped row never enters the dedup sets and never blocks its own settled copy from being imported on a later sync.
* Filtered on our side rather than through the API's `transaction_status` parameter, because not every ASPSP populates `status` and filtering server-side risks empty responses from the banks that never send it. `BOOK`, `OTHR` and a missing field import as before.

**Where the settlement logic lives**

`TransactionSettlement` owns both signals a bank can use to say "this one has not settled", because they are the same question and reading them apart invites the wrong conclusion — that either had made the other redundant:

* `status` ∈ {`PDNG`, `HOLD`, `SCHD`, `CNCL`, `RJCT`} — the new gate. Revolut, Santander and Sabadell account for 5.6k of the 7.6k `PDNG` rows in production.
* `bank_transaction_code` `MCRD`/`UPCT` → `CCRD`/`POSD` — the canonicalization from #827, moved here unchanged. N26 sends `BOOK` on both copies of every row since July and only moves the card code, so the new gate never fires for it.

Neither covers the other's banks. `TransactionFingerprint` keeps only what identifies a transaction: `UNSTABLE_ID_BANKS` and the hashing itself.

**No cleanup of existing rows**

An earlier revision shipped a migration that soft-deleted every `status = 'PDNG'` row. It was dropped: production holds 7624 of them across 122 users, but only 1099 (14%) have a settled twin at the same amount within 7 days, and even a deliberately generous match (±5% amount, ±30 days) reaches just 4216 (55%). For the remaining 3408 rows — 110 users, ~137k € net — the pending row is the only record of that money. It also bypassed `TransactionDeleted`, which would have orphaned 1034 `budget_transactions` rows across 76 budget periods.

Rows already stored are left for users to delete themselves, the same call #827 made for the N26 duplicates.

**Known trade-off**

A pending delivery whose settled copy never arrives, or arrives after the 3-day re-fetch window (`WATERMARK_OVERLAP_DAYS`), is now never imported at all. In production 45% of `PDNG` rows have no settled counterpart and 61% are already older than 30 days, so this is not hypothetical: for those banks some movements will simply not appear, with no error and no placeholder. Nothing downstream breaks — balances, net worth, budgets, categorization and `banking:health` have no concept of a pending row — but nothing surfaces the gap to the user either. Representing pending transactions instead of dropping them is the follow-up; correct totals beat double-counted ones in the meantime.

**Testing**

* `TransactionSyncServiceTest`: the five unsettled statuses are skipped (parameterized), `BOOK`/`OTHR`/absent import normally, and the real duplicate scenario — pending on one sync, settled on the next — ends with a single row carrying the settled amount and status.
* `TransactionSettlementTest`: both signals directly, including card-code pairs that half-match and must stay discriminating, and a lowercase status that must not be mistaken for an unsettled one.


victor-falcon:
We were doing this only for a few banks that we’ve found can’t give us the same ID when their status changes, so the transactions end up being duplicated.

Which bank have you seen this issue with?

victor-falcon:
The migration should come out of this PR — it soft-deletes far more than the duplicate population it targets.

Production, for `source = 'enablebanking' AND raw_data->status = 'PDNG'`:

- **7624 rows across 122 users.** Only **1099 (14%)** have a settled twin at the same amount within ±7 days. With a deliberately generous match (±5% amount, ±30 days) it is still only 4216 (55%).
- **3408 rows (45%, 110 users, ~137k€ net) have no settled counterpart at all.** For those the pending row is the only record of that money, and `down()` is a no-op.
- 61% are older than 30 days, so they are never going to settle.
- None of them are N26, so the "7916 N26 rows" the migration comment cites is not the population it touches. `HOLD` and `SCHD` have zero rows too.

Separately, the mass `update(['deleted_at' => now()])` bypasses `TransactionDeleted`, so `UnassignTransactionFromBudget` never runs — that leaves 1034 `budget_transactions` rows orphaned across 76 budget periods (18.4k€).

Suggest shipping the sync change on its own. If we do want a cleanup later, it should be restricted to rows with a demonstrable settled twin and done model-side in chunks so the budget listener fires.

michnhokn:
> We were doing this only for a few banks that we’ve found can’t give us the same ID when their status changes, so the transactions end up being duplicated.
> 
> Which bank have you seen this issue with?

Ah, I see!
For me it was the DKB.

victor-falcon:
Pushed two commits to this branch (thanks for letting maintainers edit):

- **Dropped the migration** for the reasons in the comment above.
- **Moved the status list into `TransactionFingerprint`**, next to `PENDING_CARD_CODE`/`SETTLED_CARD_CODE`, so both settlement filters read side by side. `TransactionSyncService` now calls `TransactionFingerprint::isUnsettled()`. No behaviour change — the N26 card-code canonicalization from #827 is untouched.

The docblock spells out why both filters have to stay: they read different fields and cover disjoint sets of banks. Banks that populate `status` leave `bank_transaction_code` alone (Revolut, Santander and Sabadell hold 5.6k of the 7.6k production PDNG rows), while N26 sends `BOOK` on both copies of every row since July and only moves the card code.

Green locally: pint, 40 tests, phpstan 0 errors, `crap` none above 10, `dry` at 4.39%.

One thing worth naming that this PR does not solve: for the banks above, 45% of pending rows never get a settled counterpart at all, and 61% are already older than 30 days. Skipping them stops the duplicates, but it also means those movements no longer show up. Same trade-off #827 measured for N26 (259 of 337). Fine as a first step — the ledger being correct beats it being double-counted — but representing pending transactions properly instead of dropping them is the real follow-up.

PR #845, opened 2026-08-22
feat: add OrcaRouter as a named AI provider

XiaoHuo888-hue:
## Add [OrcaRouter](https://www.orcarouter.ai) as a named AI provider

This PR adds **OrcaRouter** as a named provider for the AI features (transaction categorization, automation-rule suggestions, and report summaries), alongside the existing Gemini, Ollama, and other laravel/ai providers.

### What changed

- **`config/orcarouter.php`** — new config for the OrcaRouter gateway: `ORCAROUTER_URL` (defaults to `https://api.orcarouter.ai/v1`), `ORCAROUTER_API_KEY`, and the default model `orcarouter/auto`.
- **`app/Providers/AppServiceProvider.php`** — registers the named `orcarouter` provider via `Ai::extend()` using laravel/ai's `openai-compatible` driver (Bearer auth + custom base URL).
- **`app/Services/Ai/AiProviderResolver.php`** — small helper that maps a provider config string to a `Lab` enum case when one exists, and otherwise passes the named provider string through to laravel/ai (so named providers like `orcarouter` resolve by config key).
- **`config/ai_categorization.php`, `config/ai_suggestions.php`, `config/ai_reports.php`** — default the model to `orcarouter/auto` when the provider is `orcarouter`; per-feature `*_MODEL` vars still override.
- **`.env.example` / `README.md`** — document `ORCAROUTER_API_KEY` / `ORCAROUTER_URL` / `ORCAROUTER_MODEL` and the `AI_PROVIDER=orcarouter` switch.
- **Tests** — unit test for `AiProviderResolver`, a feature test proving `AI_CATEGORIZATION_PROVIDER=orcarouter` routes prompts through the named provider, and an updated fail-fast test (unknown providers now surface `InvalidArgumentException` from laravel/ai's manager instead of `ValueError` from `Lab::from`).

### How it works

Set `AI_PROVIDER=orcarouter` (or a per-feature `*_PROVIDER`) and `ORCAROUTER_API_KEY`:

```dotenv
AI_PROVIDER=orcarouter
ORCAROUTER_API_KEY=your-orcarouter-key
```

laravel/ai then routes all three AI features to `https://api.orcarouter.ai/v1/chat/completions` with the configured key. Gemini stays the default, so existing deployments are unaffected.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible gateway that routes to frontier models over a single endpoint. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

### Verification

- `php -l` passes on all changed files.
- Pint (`laravel` preset) passes on all changed files.
- `git diff --check` is clean.
- **L3 live test**: `POST https://api.orcarouter.ai/v1/chat/completions` with `model=orcarouter/auto` and a real key returns **HTTP 200** with a valid `chat.completion` — the exact request shape the `openai-compatible` driver sends.

Disclosure: I'm an engineer on the OrcaRouter team.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -11,8 +11,12 @@ use App\Services\Ai\UncategorizedTransactionMatcher;
 use App\Services\Banking\EnableBankingProvider;
 use App\Services\Discord\DiscordWebhook;
 use Illuminate\Cache\RateLimiting\Limit;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Support\Facades\RateLimiter;
 use Illuminate\Support\ServiceProvider;
+use Laravel\Ai\Ai;
+use Laravel\Ai\Providers\OpenAiCompatibleProvider;
 use Laravel\Cashier\Cashier;
 use Laravel\Fortify\Contracts\RegisterResponse as RegisterResponseContract;
 use Laravel\Passport\Passport;
@@ -60,5 +64,22 @@ class AppServiceProvider extends ServiceProvider
         // Render the OAuth consent screen (Claude Desktop / ChatGPT connecting
         // to the MCP server) with our own on-brand Blade view.
         Passport::authorizationView(fn (array $parameters) => response()->view('mcp.authorize', $parameters));
+
+        // Register the named OrcaRouter provider as an openai-compatible instance.
+        // Referencing it is as easy as AI_PROVIDER=orcarouter — laravel/ai routes
+        // the three AI features through this gateway. See config/orcarouter.php.
+        // Runs here (not register()) so laravel/ai's AiManager binding exists.
+        Ai::extend('orcarouter', function (Application $app, array $config): OpenAiCompatibleProvider {
+            return new OpenAiCompatibleProvider(
+                array_merge($config, [
+                    'url' => config('orcarouter.url'),
+                    'key' => config('orcarouter.key'),
+                    'models' => [
+                        'text' => ['default' => config('orcarouter.model')],
+                    ],
+                ]),
+                $app->make(Dispatcher::class),
+            );
+        });
     }
 }
--- a/config/ai_categorization.php
+++ b/config/ai_categorization.php
@@ -9,15 +9,18 @@ return [
     |
     | Provider and model for transaction categorization, both env-overridable.
     | The provider defaults to Gemini but accepts any laravel/ai provider (a
-    | valid Laravel\Ai\Enums\Lab case — an unknown value fails fast). Cost is
-    | negligible at any tier, so the model is chosen for accuracy. See the
-    | README "AI Provider" section for the shared options (e.g. local Ollama).
+    | valid Laravel\Ai\Enums\Lab case or a named provider like "orcarouter").
+    | Cost is negligible at any tier, so the model is chosen for accuracy. See
+    | the README "AI Provider" section for the shared options (e.g. local Ollama
+    | or the OrcaRouter gateway).
     |
     */
 
     'provider' => env('AI_CATEGORIZATION_PROVIDER', env('AI_PROVIDER', 'gemini')),
 
-    'model' => env('AI_CATEGORIZATION_MODEL', 'gemini-flash-latest'),
+    'model' => env('AI_CATEGORIZATION_MODEL', env('AI_CATEGORIZATION_PROVIDER', env('AI_PROVIDER', 'gemini')) === 'orcarouter'
+        ? env('ORCAROUTER_MODEL', 'orcarouter/auto')
+        : 'gemini-flash-latest'),
 
     /*
     |--------------------------------------------------------------------------
--- a/tests/Unit/Ai/AiProviderResolverTest.php
+++ b/tests/Unit/Ai/AiProviderResolverTest.php
@@ -1,0 +1,15 @@
+<?php
+
+use App\Services\Ai\AiProviderResolver;
+use Laravel\Ai\Enums\Lab;
+
+it('resolves known lab provider names to their enum case', function () {
+    expect(AiProviderResolver::resolve('gemini'))->toBe(Lab::Gemini);
+    expect(AiProviderResolver::resolve('openai'))->toBe(Lab::OpenAI);
+    expect(AiProviderResolver::resolve('openai-compatible'))->toBe(Lab::OpenAiCompatible);
+    expect(AiProviderResolver::resolve('ollama'))->toBe(Lab::Ollama);
+});
+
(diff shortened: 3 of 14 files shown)


tests/Unit/Ai/AiProviderResolverTest.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+use App\Services\Ai\AiProviderResolver;
+use Laravel\Ai\Enums\Lab;
+
+it('resolves known lab provider names to their enum case', function () {
+    expect(AiProviderResolver::resolve('gemini'))->toBe(Lab::Gemini);
+    expect(AiProviderResolver::resolve('openai'))->toBe(Lab::OpenAI);
+    expect(AiProviderResolver::resolve('openai-compatible'))->toBe(Lab::OpenAiCompatible);
```
victor-falcon: **Blocker — this test fails.** `Lab::OpenAiCompatible` doesn't exist. The pinned `laravel/ai` v0.10.3 declares the case as `OpenAICompatible` (capital `AI`) in `vendor/laravel/ai/src/Enums/Lab.php:19`, and enum case access is a case-sensitive constant lookup.

Ran against the installed vendor:

```
pest tests/Unit/Ai/AiProviderResolverTest.php
→ tests: 2, passed: 1, errors: 1
→ line 9: Undefined constant Laravel\Ai\Enums\Lab::OpenAiCompatible
```

`php -l` can't catch this — it's syntactically valid — which is why the stated verification (`php -l`, Pint, `git diff --check`) missed it. `pest` is a required check, so this reddens CI.

Fix is one letter: `Lab::OpenAICompatible`.

app/Providers/AppServiceProvider.php
```diff
@@ -60,5 +64,22 @@ public function boot(): void
...
+                array_merge($config, [
+                    'url' => config('orcarouter.url'),
+                    'key' => config('orcarouter.key'),
+                    'models' => [
+                        'text' => ['default' => config('orcarouter.model')],
+                    ],
+                ]),
+                $app->make(Dispatcher::class),
+            );
+        });
```
victor-falcon, line 83: Two things on this block.

**Nothing exercises this closure.** All 20 feature tests pass, but every one of them goes through `TransactionCategorizationAgent::fake()` / `RuleSuggestionAgent::fake()`, so `AiManager` never resolves this provider. A mistake in the `array_merge` precedence, the `models.text.default` path, or the `orcarouter.url` / `orcarouter.key` wiring would ship undetected — which is exactly the gap the enum typo in the unit test fell through.

**The package already does this declaratively.** `AiManager::createOpenaiCompatibleDriver()` (`vendor/laravel/ai/src/AiManager.php:420-426`) builds the provider straight from config, so a published `config/ai.php` with

```php
'providers' => [
    'orcarouter' => ['driver' => 'openai-compatible', 'url' => ..., 'key' => ..., 'models' => ...],
],
```

would register this with zero new PHP, and cover every other OpenAI-compatible gateway at the same time, instead of a vendor-named config file. Hand-rolling `new OpenAiCompatibleProvider(...)` couples app code to the package's constructor signature.

To be fair: the `extend()` route *does* work — `getInstanceConfig()` falls back to `['driver' => $name]` (`AiManager.php:494-498`) and `customCreators` is keyed by driver name, so `provider: 'orcarouter'` finds this closure. It's not a bug, just more machinery than the framework needs, and it diverges from how Gemini and Ollama are referenced (bare enum case, no per-provider registration).

config/ai_categorization.php
```diff
@@ -9,15 +9,18 @@
...
+    | or the OrcaRouter gateway).
     |
     */
 
     'provider' => env('AI_CATEGORIZATION_PROVIDER', env('AI_PROVIDER', 'gemini')),
 
-    'model' => env('AI_CATEGORIZATION_MODEL', 'gemini-flash-latest'),
+    'model' => env('AI_CATEGORIZATION_MODEL', env('AI_CATEGORIZATION_PROVIDER', env('AI_PROVIDER', 'gemini')) === 'orcarouter'
+        ? env('ORCAROUTER_MODEL', 'orcarouter/auto')
+        : 'gemini-flash-latest'),
```
victor-falcon, line 23: This exact ternary is triplicated — here, plus `config/ai_reports.php:22-24` and `config/ai_suggestions.php:22-24`, differing only by env prefix. It also re-derives the provider fallback chain that the line above already computed for `'provider'`, so the two can drift apart.

`bun run dry` won't flag it: `.jscpd.json` only scans `app` and `resources/js`, not `config/`. A small helper (`AiProviderResolver::defaultModelFor($provider)`) collapses all three.

Separately, and worth a deliberate decision: the special case covers only `orcarouter`. With `AI_PROVIDER=ollama` (or openai / anthropic / groq / …) the model default still falls through to `'gemini-flash-latest'` — a Gemini model against a non-Gemini provider. That footgun predates this PR, but given the disclosed affiliation, generalizing the smart default reads better than reserving it for one vendor.

victor-falcon (commented):
Review of this PR. I ran the checks the description doesn't mention, against the installed vendor at `0dc125d`:

| Check | Result |
| --- | --- |
| `pest tests/Unit/Ai/AiProviderResolverTest.php` | :x: **1 of 2 fails** |
| `pest tests/Feature/Ai/{CategorizeTransactions,RuleSuggestionGenerator}Test.php` | :white_check_mark: 20/20 |
| `pint --test` | :white_check_mark: |
| `phpstan analyse` | :white_check_mark: 0 errors |
| `artisan crap --base=origin/main` | :white_check_mark: none above complexity 10 |
| `bun run dry` | not runnable locally (binary absent) |

One blocker, and it's a one-letter fix — commented inline on the unit test. Everything else is inline too.

Verified non-issues, so nobody has to re-litigate them:

- **Fail-fast is preserved.** An unknown provider still throws before any network call, now `InvalidArgumentException("Instance driver [x] is not supported.")` instead of `ValueError`, and the updated test covers it. Confirmed by running it, not just reading it.
- **No `.env.example` trap.** The new `ORCAROUTER_*` lines are all commented out, so the "blank var reads back as `''` rather than `null` in CI" problem doesn't apply here.
- **Gemini really does stay the default**, in all three feature configs, including the case where `AI_PROVIDER` is set but a per-feature `*_PROVIDER` isn't.
- **No dependency or base-folder changes**; no new `__()` strings, so the `lang/es.json` gate is unaffected.

Thanks for disclosing the affiliation up front — that's the right call. I've flagged the two spots where it's worth a second look: the README copy and the vendor-only model default.

victor-falcon:
Connecting OrcaRouter today something you can do just with environment variables.

I’m going to close this pull request and add a simple guide to the README so it’s easy for everyone.

https://github.com/whisper-money/whisper-money/pull/851

PR #889, opened 2026-08-28
feat(money): store and show each currency at its own number of decimals

victor-falcon:
## Why

A forum user reported that a manual Bitcoin account could only be fractioned to 2 decimals, so their real holding could not be represented at all. The same truncation hit the 3-decimal fiat currencies (KWD), and in the other direction every peso, guarani and yen amount printed a meaningless `,00`.

The cause was a single assumption: **every money value was an integer in minor units at a hardcoded scale of 2 decimals**, whatever the currency. There was no per-currency scale anywhere — not in the schema, not in the config, not in the formatters.

`currency-decimals-assessment.md` (first commit) is the costed study behind this, including the production numbers and the options that were rejected.

## What changed

The scale now comes from a `decimals` key in `config/currencies.php`, shared to the frontend so both sides read one table rather than each deriving its own from CLDR. Only **7 of 36** currencies move:

| Currencies | Decimals | Accounts in prod |
| --- | --- | --- |
| COP, CLP, PYG, JPY, PKR | 2 → **0** | 696 |
| KWD | 2 → **3** | 5 |
| BTC | 2 → **8** | 8 |
| the other 29 | 2 (unchanged) | — no row touched |

### Storage

One migration rescales the 11 money columns across 7 tables. Five of those tables carry **no currency column of their own**, so each resolves it by a different path — that, not the config key, is where the cost of this change sits:

| Column | Currency resolved through |
| --- | --- |
| `transactions.amount` | its own `currency_code` |
| `account_balances.balance`, `.invested_amount` | the account |
| `real_estate_details.purchase_price`, `loan_details.original_amount` | the account |
| `budget_transactions.amount` | the **transaction** it snapshots (copied without conversion) |
| `budget_periods.allocated_amount`, `.carried_over_amount` | the owner's primary currency |
| `savings_goals.target_amount`, `.initial_amount`, `.archived_saved_amount` | the owner's primary currency |

`savings_goals.archived_saved_amount` is widened from `int` to `bigint`, which a 3- or 8-decimal currency would have overflowed at 2.1e9.

The migration **hardcodes the scales** instead of reading the config: a data migration is one fixed transition between two known states and must not rescale a currency added to the config years later. It also leaves `updated_at` untouched, so it does not force every offline client to re-pull its entire history — the client drops the now-wrongly-scaled cache through a Dexie version bump instead.

### Display and input

`formatCurrency` and `AmountDisplay` stopped forcing 2 decimals, so ~140 call sites inherit the right precision; the ~20 that pass `0` deliberately for compact chart labels keep working. `AmountInput` had five values pinned to 2 — its formatter, its parser, the focus round-trip, the inline calculator and the placeholder — and now follows the currency.

## Two bugs this surfaced, both fixed here

1. **`ExchangeRateService::convert` applied the rate straight to the stored integers.** That is only correct when source and target share a scale; it was off by a power of ten between any two unequal ones. `100.000 COP` came out as **0,23 €** instead of 23,26 €. This is the single point every cross-currency total goes through — dashboard, cashflow, net worth — so it affected all of them. It now converts in major units and rescales to the target.
2. **`AmountInput` read `1.234.567` as `1`**, so entering a million pesos was impossible. Every separator in a zero-decimal currency is now grouping; a repeated separator is grouping in any currency (in euros, `1.234.567` used to become 1,23 €).

## One simplification

`LoanAmortizationService` lost its six conversions. The amortization formulas are homogeneous in the principal, so dividing by 100 and multiplying back was a no-op that only cost precision. It is now scale-agnostic and its 10 tests pass unchanged.

## Accepted losses

- **Scaling down to 0 decimals is lossy, and that was the explicit call.** 1,874 COP transactions and 1,542 COP balance rows carry non-zero centavos and lose them. `down()` restores the scale, never those centavos. **Take a backup before deploying.**
- **BTC values already truncated to 2 decimals cannot be recovered by any migration.** The 7 affected users are padded with zeros (`0.80` → `0.80000000`) and need to re-enter their real balances. That needs an email, not code.
- **The migration and the frontend must deploy together.** Frontend first shows COP amounts divided by 100 for the length of the window; migration first, multiplied by 100. Minutes of wrong numbers, no data at risk.

## Guardrails

A new test compares the config against CLDR for all 36 currencies and fails on drift, with BTC as the one declared override (it is not an ISO currency, so CLDR gives it 2). That is what stops the table rotting, and what would catch a currency added later without its scale.

## Testing

- **2790 PHP tests pass** (1 skipped, 1 incomplete, both pre-existing) and **617 JS tests**.
- New: 9 migration tests (rounding direction, nulls preserved, `updated_at` untouched, the widened column, `down()`, and that a budget transaction follows the transaction's currency and not the owner's), the CLDR drift test, per-scale `Money` and `formatCurrency` tests, `AmountInput` tests for COP/BTC/EUR entry and the separator cases, and cross-scale `ExchangeRateService` tests.
- `pint`, `eslint`, `prettier`, `jscpd` and `bun run build` all clean.

Note for reviewers running the suite locally: if you have run `bun run build` in your worktree, six Inertia page tests will fail with `409`. That is the Vite manifest making `version()` non-null while those tests send no `X-Inertia-Version`; CI runs the tests without the build artifact. Move `public/build` aside and they pass.

### Manual QA

Verified in the browser against the real app, checking the database and not just the pixels:

| Case | On screen | In the database |
| --- | --- | --- |
| COP balance | `1.234.567 COP` | `1234567` |
| Typing `1.234.567` in COP | accepted as typed | `1234567` |
| COP transaction | `25.000 COP` | `25000` |
| **BTC balance `0.00123456`** | `0,00123456 BTC` | `123456` satoshis |
| USD regression | `1.234,56 US$` | `123456` cents |

## Demo

**📎 Attach video here** — recorded at `~/Downloads/currency-decimals-qa.mp4`

https://github.com/user-attachments/assets/5c203dfd-b761-4eec-9574-ef98f1d1eaef

## Not in this PR

- **BHD and TND** are not in `config/currencies.php` at all, so they cannot be selected. The reporter named them alongside KWD; they are not fixed, just unavailable. The CLDR test would catch them if added without a scale.
- `currency_code` is `varchar(3)`, so BTC fits but USDT and USDC would not. Stated limit, not widened here.
- The symbol tables (`Money::format`'s 5-entry `match`, `getCurrencySymbol`'s 10-entry map) cover 15 of 36 currencies. `Intl.NumberFormat(...).formatToParts()` — already used correctly in `AmountInput` — replaces both. Separate concern.
- **A pre-existing inconsistency worth its own issue:** `IndexaCapitalBalanceSyncService` writes `balance` in the account's currency but `invested_amount` in the user's. One account in production (EUR/USD, both 2 decimals), so this migration does not corrupt it, but the two columns of one row are in different currencies.
- 66 money rows already exceed `Number.MAX_SAFE_INTEGER` and are cast to `integer`, so the browser receives corrupted values today. Independent of scale; needs an upper bound on amount input.

### DIFF
--- a/app/Console/Commands/SendDailyStatsReportCommand.php
+++ b/app/Console/Commands/SendDailyStatsReportCommand.php
@@ -123,11 +123,11 @@ class SendDailyStatsReportCommand extends Command
     }
 
     /**
-     * MRR figures are held in major units (e.g. euros), so convert to cents for
-     * the shared formatter.
+     * MRR figures are held in major units (e.g. euros), so convert to the
+     * currency's minor units for the shared formatter.
      */
     private function money(float $amount, string $currency): string
     {
-        return Money::format((int) round($amount * 100), $currency);
+        return Money::format(Money::toMinor($amount, $currency), $currency);
     }
 }
--- a/app/Console/Commands/StripeSubscriptionStatsCommand.php
+++ b/app/Console/Commands/StripeSubscriptionStatsCommand.php
@@ -79,11 +79,11 @@ class StripeSubscriptionStatsCommand extends Command
     }
 
     /**
-     * MRR figures are held in major units (e.g. euros), so convert to cents for
-     * the shared formatter.
+     * MRR figures are held in major units (e.g. euros), so convert to the
+     * currency's minor units for the shared formatter.
      */
     private function format(float $amount, string $currency): string
     {
-        return Money::format((int) round($amount * 100), $currency);
+        return Money::format(Money::toMinor($amount, $currency), $currency);
     }
 }
--- a/app/Console/Commands/SyncStripePricesCommand.php
+++ b/app/Console/Commands/SyncStripePricesCommand.php
@@ -73,6 +73,8 @@ class SyncStripePricesCommand extends Command
             return 'skipped';
         }
 
+        // Stripe's own smallest-currency-unit table, not the app's per-currency
+        // scale: this value goes out to the Stripe API, not into our columns.
         $amountInCents = (int) round($plan['price'] * 100);
         $billingPeriod = $plan['billing_period'] ?? null;
         $productId = config('subscriptions.products.pro');
--- a/app/Http/Middleware/HandleInertiaRequests.php
+++ b/app/Http/Middleware/HandleInertiaRequests.php
@@ -115,6 +115,10 @@ class HandleInertiaRequests extends Middleware
             'currencies' => [
                 'profile' => $this->currencyOptions->primaryOptions(),
                 'accounts' => $this->currencyOptions->accountOptions(),
+                // Keyed by code rather than folded into the two lists above:
+                // a transaction can be in a currency no longer offered, and it
+                // still has to render at the right scale.
+                'decimals' => $this->currencyOptions->decimalsMap(),
             ],
         ];
     }
--- a/app/Mcp/Servers/WhisperMoneyServer.php
+++ b/app/Mcp/Servers/WhisperMoneyServer.php
@@ -46,7 +46,10 @@ Access to the authenticated user's Whisper Money finance data, for analysing
 spending, cashflow and net worth — and, with write access, for editing that
 data.
 
-- All amounts are integers in minor units (cents). Divide by 100 for a display value.
+- All amounts are integers in the minor units of their own currency, and how many
+  those are per major unit depends on the currency: 100 for EUR and USD, 1 for
+  COP, CLP, PYG, JPY and PKR, 1000 for KWD, 100000000 for BTC. Read the row's
+  `currency` before scaling one, and never assume cents.
 - Data is organised into "spaces" (the personal space and any shared spaces).
   Transaction, account, category and label tools accept an optional `space` id and
   default to the personal space; call `list_spaces` to discover ids. The cashflow,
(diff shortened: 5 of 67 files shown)


victor-falcon:
The `crap` check is red (not a required check). It reports three methods above complexity 10:

- `BinanceBalanceSyncService::syncHistoricalBalances`
- `IndexaCapitalBalanceSyncService::sync`
- `InteractiveBrokersBalanceSyncService::sync`

All three were already at complexity 12 before this branch. `crap --base=origin/main` lists them because the diff *touches* them, and the entire change in each is swapping one expression for another:

```diff
-        return (int) round($totalValue * 100);
+        return Money::toMinor($totalValue, $targetCurrency);
```

Not one branch added, so the cyclomatic complexity is unchanged.

I deliberately did neither of the two things that would turn it green:

- **Not refactored.** Splitting up three bank-sync methods against live external APIs is a real risk and unrelated to per-currency decimals. It belongs in its own PR.
- **Not added to `.crap-ignore.json`.** That file is for complexity that is genuinely warranted; using it here would file pre-existing debt as justified and hide it.

Leaving it red reports the truth: those methods need attention, and not because of this change.

PR #909, opened 2026-09-02
fix(summaries): give the card renderer a writable HOME

victor-falcon:
Fixes PHP-LARAVEL-5M.

## The bug

Every shareable card rendered on demand failed in production. Chromium died during launch:

```
[pid=578][err] chrome_crashpad_handler: --database is required
[pid=578] <process did exit: exitCode=null, signal=SIGTRAP>
error: launch: Target page, context or browser has been closed
```

Chromium resolves its crash-dump database under `HOME` and refuses to start when it cannot. The evidence for that being the missing piece is already in this repo — `docker/supervisor/supervisord.conf` gives `HOME="/tmp"` to `queue-worker-emails` and says why in a comment.

That explains the split:

| format | rendered by | `HOME` | result |
| --- | --- | --- | --- |
| `feed` | `queue-worker-emails`, up front | `/tmp` | works |
| `story`, `wide` | `MonthlySummaryController@card`, on first request | php-fpm has no `environment=` line | 503, every time |

So the card in the email was fine and every "download" click on the other two formats returned 503.

`PLAYWRIGHT_BROWSERS_PATH` was never the problem: the failing production log shows Chromium being found at `/usr/local/playwright/chromium-1208/`, because `Dockerfile.production` sets that one container-wide. Only `HOME` was missing.

## The fix

`->env(['HOME' => sys_get_temp_dir()])` on the render process. Laravel merges onto the inherited environment rather than replacing it, so everything else the child needs is untouched.

Setting it here rather than adding a second `environment=` line to `[program:php-fpm]` keeps the requirement with the code that has it — it now holds from a worker, a web request, the scheduler, the preview command, or whatever calls it next. Giving the whole FPM pool a `HOME` to fix one subprocess is a wider change than the problem.

## What I deliberately did not do

**Removed `HOME="/tmp"` from `queue-worker-emails`.** It is redundant for cards now, and the review argued for deleting it. It is one env var, and nothing has established that the rest of that queue doesn't want a writable home — so deleting it to tidy up risks a production queue for no functional gain. The misleading half was the comment, which claimed that line is what makes cards work; that is what made php-fpm's missing copy so easy to overlook, so the comment now says where the real reason lives and when the line can go.

**Scoped `HOME` to a per-render directory** so Chromium's crashpad artifacts get cleaned up with the temp files. Playwright already hands the browser its own `--user-data-dir` and cleans that up; what is left under `HOME` is a small, bounded crash-dump directory in a container's `/tmp`. Creating and recursively deleting a directory per render costs more than it saves.

## Tests

`CardRenderer` had no direct coverage — every other test in the suite mocks it. This adds a test file that covers:

- the writable `HOME` reaching the process (red before this fix: *"An expected process was not invoked"*)
- the error output being surfaced in the exception message
- the happy path: the PNG landing on the public disk, the `exists()` short-circuit on the second ask, and the Pro badge being part of the cache key rather than overwriting the unbadged file

## Verification, and its limit

Red confirmed before the fix, green after. `tests/Feature/MonthlySummary` + the preview command: 82 passed. `pint --test`, `crap --base=origin/main` and `bun run dry` clean.

Worth being straight about the ceiling: the test fakes the process, so it proves the plumbing, not that a real Chromium starts. The diagnosis is inference from the crashpad error, the fpm-vs-queue split and this repo's own config comment — strong, but not executed against production. The change is additive and cannot break the path that already worked, so the downside of being wrong is that the bug persists rather than anything new breaking. A click on "download story" after deploy is the real check.

Nothing to clean up for affected users: `render()` only writes to the public disk after a successful screenshot, so no broken file was ever cached. Their next click renders.

## Follow-up worth a ticket, not this PR

A card render launches a cold Chromium with a 60s timeout synchronously inside a web request, and the failure path is a bare `abort(503)` with no product copy or retry. Both predate this bug.

### DIFF
--- a/app/Services/MonthlySummary/CardRenderer.php
+++ b/app/Services/MonthlySummary/CardRenderer.php
@@ -22,6 +22,11 @@ use RuntimeException;
  * the one that rides inside the email and backs the public page's og:image. The
  * other two are rendered the first time somebody asks for them and then cached,
  * so nobody pays for twelve images a user will never open.
+ *
+ * Chromium needs a writable HOME to start, and it is started from two very
+ * different places — a queue worker for the feed card, a web request for the
+ * other two — so the requirement travels with the renderer rather than with
+ * whichever process manager happens to be calling it.
  */
 class CardRenderer
 {
@@ -102,14 +107,16 @@ class CardRenderer
 
     private function screenshot(string $htmlFile, string $pngFile, int $width, int $height): void
     {
-        $result = Process::timeout(self::TIMEOUT_SECONDS)->run([
-            (string) config('monthly_summary.node_binary'),
-            base_path('scripts/render-card.mjs'),
-            $htmlFile,
-            $pngFile,
-            (string) $width,
-            (string) $height,
-        ]);
+        $result = Process::timeout(self::TIMEOUT_SECONDS)
+            ->env(['HOME' => sys_get_temp_dir()])
+            ->run([
+                (string) config('monthly_summary.node_binary'),
+                base_path('scripts/render-card.mjs'),
+                $htmlFile,
+                $pngFile,
+                (string) $width,
+                (string) $height,
+            ]);
 
         if ($result->failed() || ! is_file($pngFile)) {
             throw new RuntimeException('Card render failed: '.trim($result->errorOutput() ?: $result->output()));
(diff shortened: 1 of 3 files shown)


victor-falcon:
Superseded by #910, which fixes the same root cause at `[supervisord]` level (covering every process in the image, not just the renderer) and goes further: cards pre-rendered in the send job, batched into one browser, and old months cleaned up.

PR #925, opened 2026-09-03
feat(achievements): 59 medals, dated when they really happened

victor-falcon:
> Stacked on #924 (the notification centre). Review that one first; this PR's diff is against it.

## What

Fifty-nine achievements, a daily sweep that finds them, a progress screen that shows every one of them — earned or not — and a card for the ones worth posting.

## The idea the whole thing rests on

A milestone is dated **when it actually happened**, not when we noticed. The sweep reads a user's entire history and reconstructs the month each threshold was crossed, so somebody who passed 50k in March 2023 gets a March 2023 medal. That is why the screen is worth opening: it is a financial life spread over years, not a grid of today.

Streaks are the deliberate exception. A streak is a habit kept while the medals were there to keep it for, so "months in a row" counts only from the launch month on. The track says so in its own note, because four empty slots next to a live seven-month streak would otherwise read as a bug.

**Nothing is ever revoked.** A medal records that something happened. The balance can fall and the streak can break; the row stays. The live streak is shown separately in the overview, because that one *is* a statement about now.

## Showing up counts too

Nine of the fifty-nine are about the habit rather than the money: five for days in a row opening the app, four for weeks in a row. They are the only track a reader can unlock on their first day, which is the point — somebody who has recorded nothing yet is still showing up, and the evaluator runs the visit track even when the history is empty.

There is no log of visits to count from. `last_active_at` is a single timestamp, overwritten on every visit, so both runs are carried forward one step at a time by the middleware that already writes that column, and both are kept on the user row — the current run and the longest. The longest is what the medal reads, because the sweep is nightly and a run that peaked and broke between two sweeps still happened. Two details the middleware had to learn: a new day is written however recent the last write (the throttle saves a write per request, but the day boundary is what a streak counts), and the boundary is the reader's own — `isSameDay` in UTC would end a Madrid reader's day at two in the afternoon.

A run is dated to the day it reached its length rather than to the first of the month, which every other medal uses: a streak that ended on the 20th did not happen on the 1st.

## Thresholds per currency

Money ladders are defined per currency in `config/achievements.php` for the seven that cover 94% of members (USD, EUR, MXN, COP, ARS, PEN, CLP), with everyone else measured in USD at the historical rate. They are scaled to **what a salary buys locally, not to the exchange rate** — at the rate, an "epic" saving month in Mexico would be one nobody ever has, and a medal nobody can reach is dead weight on the page.

Thresholds are stored in major units and scaled to the currency's own minor units on read, so COP (no centavos) and EUR (cents) are compared against stored integers correctly.

## The screen

Fifty-nine medals, most of them locked for most people, is a wall of failure drawn as a grid. It is drawn as thirteen ladders instead: each track runs small to large, what has been earned sits left, and the empty slots trail off to the right where they read as the road ahead. A Timeline view shows the same medals by year.

- **Locked medals send nothing but a tier.** No name, no pictogram, no threshold: a silhouette that leaks its own name through the network tab is not a silhouette.
- **Rarity without colour.** The palette is monochrome, so the tier is the shape of the rim (disc, ringed disc, rosette, rosette with an inner ring) and the weight of its label. The real share of members who hold a medal is shown too, but only above a floor of evaluated members.
- **Figures travel as numbers.** `{type, value, currency}` rather than a formatted string, so amounts go through `AmountDisplay` and privacy mode blanks them like everywhere else. Days and weeks are figure types of their own, singular included, so "1 day" reads like "1 month".

## Posting a medal

A medal is drawn server-side by the same headless browser the monthly summary card uses: two shapes (4:5 for a feed, 9:16 for a story), both skins, in the reader's language, handed to the device's own share sheet with saving the picture as the fallback. A medal can withhold its amount, which a summary card cannot — that is the one thing the two dialogs disagree about.

Everything else they share, in `ShareCardDialog`: the shapes offered, the preview that warms the render before the share gesture, and the share-or-save pair. The monthly summary screen now opens that same dialog instead of three download links and a light/dark switch, so the two cannot drift. 16:9 is gone from both — nothing ever unfurled it.

## Telling people

One row in the bell per medal, and one grouped email per sweep day — three medals on a Tuesday is three rows and one message. The first sweep is different: it reads a whole history at once and can unlock twenty medals in a second, so those are recorded silently and a single welcome row says how many. The email is opt-out (`notify_achievements`, with a one-click unsubscribe); the row is not, because it is the record of what happened, not a message.

The toggle for that email is left out of the preferences screen entirely while the medals are switched off for the reader, rather than shown disabled: a toggle for an email nothing would ever send is worse than no toggle.

## Decisions worth arguing with

- **The sweep reads the whole history every night**, not just the months since last time. Someone who imports three years of statements today changes what happened in 2023, and a medal dated by when we noticed would be a lie. If it stops being cheap, the fix is to skip users with no new transactions — not to shorten the window.
- **The first savings goal is dated when the goal finished**, not by the contribution that crossed the line. Nothing records the crossing and reconstructing it means walking every tagged transaction of every goal, for one medal.
- **The unsubscribe controller was generalised** rather than copied for a second preference. Existing signed links keep working: same URL, same route name.
- **The demo account is swept like anyone else** (its emails are already switched off by `demo:reset`), so a freshly reset demo has a full progress screen by morning.
- **Waiting for a preview is tracked as which URL has painted**, not as a flag reset by hand. That also fixes a real bug: the picture is cached per language, but the URL was the same in every language, so after switching language the browser answered from its own cache with the old one.

## Not in this PR

- **The dashboard achievement strip**, which the brief marked optional.

## Testing

- `AchievementCatalogTest` — the catalog's shape, and that every ladder is complete and ascending in every currency it ships.
- `AchievementEvaluatorTest` — the rules, against a history handed straight in: which medals a given past contains and, just as much, the month each is dated to.
- `AchievementSweepTest` — end to end from real transactions: dates, the backfill being announced once, a medal written once however often the sweep runs, a medal surviving the deletion of the money that earned it, and the fallback currency.
- `AchievementScreenTest` — every medal sent, locked ones keeping their names to themselves, figures as numbers, the ladder in the reader's currency, and the rarity floor.
- `AchievementCardTest` and `Cards/CardBrowserTest` — the card's own route and presenter, both shapes and skins, the language it is drawn in, and the browser the render subprocess ends up using.
- `TrackLastActiveAtTest` — both runs: the day extended, the day missed, a second visit the same day, a minute after midnight, the reader's own timezone, and the same four for weeks.
- Vitest covers the medal, its palette, the figures (days, weeks and months in the singular) and both share dialogs.

Behind `ACHIEVEMENTS_ENABLED` (Pennant, off by default).


victor-falcon:
CI note: this workflow only runs for pull requests targeting `main` or `develop`, so a stacked PR gets no checks of its own. A run was dispatched manually against this branch: https://github.com/whisper-money/whisper-money/actions/runs/33775560548

Once #924 merges, GitHub retargets this PR to `main` and the usual checks run on it.

PR #965, opened 2026-09-10
fix(achievements): bound the sweep to the history it can value and to one reader's memory

victor-falcon:
Fixes PHP-LARAVEL-5Z
Fixes PHP-LARAVEL-61

The first production `achievements:sweep` after cf7e252a took the Pennant gate
off met the real member base and threw twice in the same run. Both failures are
the same shape — a reader with a very long transaction history — and neither is
reachable from the seeded accounts the job had been running against until then.

## PHP-LARAVEL-5Z — a permanent gap treated as a transient one

`HistoryBuilder::for()` read every reader from their earliest
`transactions.transaction_date` with nothing bounding how far back that is, and
`ensureRates()` refuses to build a history when any month end lacks a rate for a
foreign-currency account. That refusal is right and stays: a medal is written
once and never revoked, so a rate that failed to arrive would date a milestone to
the wrong month, or award one on a figure nobody ever had, permanently. Better to
record nothing today and let tomorrow's sweep try again.

Only there is no tomorrow for January 2001. The rate provider holds nothing
before **2024-03-04** — verified by probing it, both the jsdelivr CDN and the
pages.dev mirror 404 for every earlier date — so five readers were skipped with a
fresh Sentry error every night, and would have been forever.

The window now starts at `achievements.rates_from` (default `2024-03`,
env-overridable) whenever the reader keeps money in a currency other than the one
their medals are measured in. `ensureRates()` is untouched for the months inside
that window, where a missing rate really is one that failed to arrive today, so
the invariant and the retry it protects both survive.

The default was checked against every base the sweep can ask for, not just one:
all seven ladder currencies and the USD fallback return a full map (~700
currencies, `btc`/`gbp`/`eur`/`usd` among them) from 2024-03-04 on, so no reader
is left in a gap between the floor and their own currency's coverage. 2024-03-31
is the first usable month end — February's 2024-02-29 is a 404 and the provider's
seven-day walk-back only goes backwards, so it never reaches March.

### The trade-off, and why this one

A floored reader's medals only reach back to the covered period: their "first
transaction" is dated to their first covered month, and the running counts
(`transactions.N`, `yearly_saving.N`) start from the floor rather than from their
true first transaction. Those figures are real and honestly computed, just
partial. Nothing is fabricated — `BalanceLookup` carries the opening balance
forward from before the range, so `net_worth.N` and the emergency-fund medals are
still dated against the balance the reader actually had.

The two alternatives were weighed and dropped:

- **Skip the uncovered months instead of trimming the window.** This is what the
  `ensureRates()` docblock forbids. A hole in the series joins two runs into one
  (`categorized` counts across the whole history), and a crossing that happened
  inside the hole gets dated to the next visible month — the same misdating,
  plus a run nobody kept.
- **Tell "the provider has no data" apart from "the lookup failed today" at the
  service.** That is the honest discriminator, but `CurrencyConversionService`
  answers a 404 and a connection failure with the same empty array, so it means
  threading a new signal up through a service ten other callers share. A date
  before the provider's coverage *is* that discriminator, and it is a date
  comparison.

Only readers holding a foreign currency are affected; a reader whose money is all
in one currency needs no rate and keeps their whole past, unchanged.

## PHP-LARAVEL-61 — the sweep's rate cache outlived the reader

`ExchangeRateService` memoizes the full rate payload per base+date for the life of
the instance, and `SweepAchievementsCommand` took `Awarder` in its **constructor**
— so the container built `Awarder → Evaluator → HistoryBuilder →
ExchangeRateService` once and every reader in the run poured their month ends into
it. Nothing evicted anything between readers, so memory grew monotonically across
the whole run until the 128 MB limit went. The `Illuminate\Http\Client\Response`
culprit frame is just where the final allocation happened to land.

The command resolves an `Awarder` per reader instead, so a reader's history is
freed with the graph that read it and the peak is one member rather than all of
them. That is the whole fix: the lifetime was the bug, and the constructor was the
lifetime. It also avoids putting a cache-eviction method on a service that
`NetWorthCalculator`, `CashflowSummaryService`, `SummaryBuilder`,
`ManualBalanceAdjuster` and five API controllers share — none of whose
`getRates()`/`convert()` behaviour changes here.

Worth naming: month-end rates are no longer shared between readers in one run, so
each foreign-currency reader now issues its own `exchange_rates` preload. That is
one batched query per reader, bounded by the window the floor gives it, in
exchange for a run that fits in memory.

## Known limitations, deliberately left

- **`safety.1` (loan paid off) can be missed by a floored reader.** It is an event
  detector: it needs to see the loan owing money and then reaching zero *inside*
  the visible months. A foreign-currency reader whose loan was cleared before the
  floor never shows as having owed anything, so the medal is not written. Dating
  it would mean either a wrong month or a second, unbounded balance walk for one
  row on a nightly job — and an unearned medal omitted is a smaller wrong than a
  medal dated to a month it did not happen in. Today those readers get no medals
  at all, so this is still strictly better for them.
- **The floor keys off any account in another currency, not off money that needs
  converting.** A stray zero-balance foreign account keeps a reader floored even
  though no rate is needed for a balance of zero. Narrowing it means modelling
  both per-month balances and per-month transaction currencies, and getting that
  subtly wrong writes a permanent row — the coarse check only ever truncates,
  never fabricates, so it stays coarse.

## Tests

`tests/Feature/Achievements/AchievementSweepTest.php`, both failing without the
fix (verified by reverting the two app files and re-running):

- `reads a foreign-currency reader from the first month a rate can exist` — a
  reader with a 2001 statement and a BTC account against a provider fake that
  answers the way the real one does. Fails with `Output "Skipped" was printed`,
  which is exactly PHP-LARAVEL-5Z.
- `reads the rates afresh for every reader instead of hoarding them for the run` —
  two readers sharing the same month ends, with the rates pre-seeded so a lookup
  is a query. Two readers must cost two readers' worth of lookups; before the fix
  the second inherits the first's cache and costs none.

The existing `records nothing when a foreign balance cannot be converted` still
passes: its months are all inside the covered window, so a provider outage still
skips the reader and retries tomorrow.

## Not in this PR

PHP-LARAVEL-60 (the `percent` column overflow) comes from the same run and is
being fixed separately. `Unlock.php` and `Evaluator::firstYearOfGrowth()` are
untouched here.

## QA

Run against an ephemeral MySQL, driving the command the way cron does.

**A reader shaped like PHP-LARAVEL-5Z** — a BTC account and monthly transactions
from 2001-01 to 2026-08 (~308 months), against a provider fake that answers the
way the real one does:

```
Swept 1 user(s), awarded 17 achievement(s).
```

No `Skipped`, 17 rows written, none dated before the floor: `transactions.1`,
`hygiene.1`, `categorized.1`, `monthly_saving.1/2`, `yearly_saving.1/2/3` and
`savings_rate.1-4` at `2024-03`, then `categorized.2-5` and `transactions.2`
progressing through 2024-05 → 2026-03 as the history accumulates. Previously: no
rows at all, and a Sentry event.

**A run over many readers, shaped like PHP-LARAVEL-61.** Readers holding money in
a currency that has to be converted, on many distinct days — because the rate
cache is keyed by base **and date**, so one full rate map is read and kept per
date, and transaction-date granularity is where a run grows rather than month
ends. Rates seeded into `exchange_rates`, and payloads sized the way the provider
really sends them (~200 currencies), so what is measured is the sweep's own
retention and not a test double's.

| 8 readers x distinct dates each | retained after the run | peak |
| --- | --- | --- |
| 100 (1,033 rate maps) - before | 16.0 MB | 173 MB |
| 100 (1,033 rate maps) - after | **4.0 MB** | 159 MB |
| 300 (2,289 rate maps) - before | 40.0 MB | 197 MB |
| 300 (2,289 rate maps) - after | **6.0 MB** | 163 MB |

That is the shape the fix is meant to change. Before, retention tracks every
reader in the run summed together (16 -> 40 MB as the run grows); after, it
tracks only the largest single reader (4 -> 6 MB), because each reader's graph is
freed with the reader. At roughly 5 MB per reader of this shape the old command
reaches 128 MB somewhere around two dozen of them, which is what the member base
handed it the first night it ran ungated.

## Checks

- `vendor/bin/pint --test` clean
- `vendor/bin/phpstan analyse` clean
- `php artisan test --exclude-testsuite=Browser --filter=Achievement` green
- `php artisan crap --base=origin/main --no-coverage` — 5 methods analysed, none
  above complexity 10
- `bun run dry` green

### DIFF
--- a/app/Console/Commands/SweepAchievementsCommand.php
+++ b/app/Console/Commands/SweepAchievementsCommand.php
@@ -28,11 +28,6 @@ class SweepAchievementsCommand extends Command
 
     protected $description = 'Record the achievements every user has earned';
 
-    public function __construct(private Awarder $awarder)
-    {
-        parent::__construct();
-    }
-
     public function handle(): int
     {
         $swept = 0;
@@ -50,10 +45,23 @@ class SweepAchievementsCommand extends Command
         return self::SUCCESS;
     }
 
+    /**
+     * One reader, with an {@see Awarder} of their own.
+     *
+     * Resolved here rather than held for the run because of what hangs off it:
+     * the rate service behind the calculators memoizes every month-end rate it
+     * reads, and one kept for the whole command grows by a reader's worth of
+     * rate maps on every pass until the process runs out of memory. A reader's
+     * history is freed with the graph that read it, so the peak is one member
+     * rather than all of them, and the cost is a dozen constructor calls
+     * against the hundreds of queries a history already takes.
+     */
     private function sweep(User $user): int
     {
         try {
-            return $this->awarder->sweep($user, notify: ! $this->option('quiet-notifications'))->count();
+            return $this->laravel->make(Awarder::class)
+                ->sweep($user, notify: ! $this->option('quiet-notifications'))
+                ->count();
         } catch (Throwable $exception) {
             report($exception);
             $this->warn("Skipped {$user->email}: {$exception->getMessage()}");
--- a/app/Services/Achievements/HistoryBuilder.php
+++ b/app/Services/Achievements/HistoryBuilder.php
@@ -12,6 +12,7 @@ use App\Services\ExchangeRateService;
 use App\Services\NetWorthCalculator;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Support\Collection as BaseCollection;
 use RuntimeException;
 
 /**
@@ -23,6 +24,10 @@ use RuntimeException;
  * hundred rows and one balance walk per user, on a nightly job — if it ever
  * stops being cheap, the fix is to skip users with no new transactions since
  * their last sweep, not to shorten the window.
+ *
+ * The whole past as far back as it can be valued, that is: a reader holding a
+ * foreign currency is read from {@see firstConvertibleMonth()} on, because a
+ * month whose money no rate can convert is a month with no figures in it.
  */
 class HistoryBuilder
 {
@@ -36,14 +41,14 @@ class HistoryBuilder
     public function for(User $user): History
     {
         $currency = $this->ladders->currencyFor($user->currency_code);
-        $first = $this->firstMonth($user);
+        $accounts = Account::query()->where('user_id', $user->id)->get();
+        $first = $this->firstConvertibleMonth($user, $accounts, $currency);
         $last = now()->subMonth()->startOfMonth();
 
         if ($first === null || $first->gt($last)) {
             return new History($currency);
         }
 
-        $accounts = Account::query()->where('user_id', $user->id)->get();
         $lookup = BalanceLookup::forAccounts($accounts->pluck('id'), $first, $last->copy()->endOfMonth());
         $months = $this->cashflow->forMonths($user->id, $currency, $first, $last->copy()->endOfMonth());
         $counts = $this->transactionCounts($user, $first, $last);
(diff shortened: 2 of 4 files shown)


victor-falcon:
Extra evidence from the Sentry event stream of that same run, gathered after this PR
was opened — it sharpens the trade-off this PR asks you to weigh.

**All five PHP-LARAVEL-5Z events, not just the 2001 one:**

| Missing rate | Month end |
| --- | --- |
| `eur` against `USD` | 2001-01-31 |
| `cad` against `USD` | 2021-07-31 |
| `gbp` against `EUR` | 2023-01-31 |
| `usd` against `EUR` | 2024-01-31 |
| `gbp` against `USD` | 2024-01-31 |

Every one falls before the `rates_from = 2024-03` floor, so the floor does cover all
five observed readers — but the cost is not uniform: two of them lose about two months
of history, the CAD reader loses three years, and the 2001 reader loses everything
before 2024. That spread is the thing to decide on, and it wasn't visible from the
single 2001 event the PR was written against.

**The memory half is more urgent than the issue counts suggest.** In the same run,
`PHP-LARAVEL-4` fired at 02:57:27 — the same second as the `FatalError` in
PHP-LARAVEL-61:

    Exception: Scheduled command ['/usr/local/bin/php' 'artisan' achievements:sweep]
    failed with exit code [255].

So the memory exhaustion does not just skip one reader: it takes the whole scheduled
sweep down, and every user still unprocessed in the `chunkById` walk gets nothing that
night. The cache-lifetime fix in this PR is what stops a nightly job that currently
dies partway through.

_Posted by the autonomous Sentry triage loop. PHP-LARAVEL-5X (#963) and
PHP-LARAVEL-60 (#964) are merged and resolved in Sentry; this PR is the one left
needing a human call._

PR #990, opened 2026-09-14
feat(onboarding): step 6 becomes an accounts hub you come back to

victor-falcon:
> Stacked on `onboarding-step-machine` (PR C). **Base is that branch, not `main`.** Nothing here is merged or rebased onto anything else.

## Why

`create-account` was a fork: Connected or Manual, picked once, step over. That is wrong about the product. Open banking returns current accounts and cards — it returns **no pension, no broker, no mortgage**. Almost everyone needs a mix: the bank for the day-to-day, a couple added by hand or by token for the rest.

Step 6 is now a hub you come back to. Three routes leave it, all three return, and the step ends when the user says they are done rather than when they have done one thing.

## The two states

| Empty | With accounts |
|---|---|
|  |  |

Captured by `tests/Browser/OnboardingAccountsHubScreenshotsTest.php` at 430×932, same convention as #981.

The second state is the one that does the work. The list is the receipt; **"Usually missed"** is the point — a mortgage or a loan (by hand, no bank shares these), a pension or a broker (Indexa and Coinbase connect with a token), another bank. Nobody adds a mortgage unprompted, and that is where most of their money is.

It is also what a resumed signup and a return from the bank land on, so it is the ordinary state from the second account onwards, not an edge case.

## What is *not* here

The three routes are unchanged and still hang where they were — PRs E, F and G rebuild them. This PR is the hub and only the hub.

## Notable

- **Progress bar.** The hub and everything behind it are one step. Both ways out are modes of this component, not steps, so the bar stays on 6/11 — the same arrangement `SUB_STEPS` already gives `import-transactions` and `import-balances`. No new mechanism.
- **`step-create-account.tsx` splits in two.** `step-accounts-hub.tsx` is the hub; `step-manual-account.tsx` is the by-hand form, moved verbatim so the hub file is the hub. E/F/G will replace the latter wholesale. The three jscpd clones it shares with `create-account-dialog.tsx` are pre-existing and moved with the code.
- **Accounts merge by id.** Created and existing accounts used to exclude each other. The step polls for accounts finalized in another browser while the user adds more here, so the same account arrives down both routes — the old either/or hid the existing ones the moment a new one was created. Covered by a test.
- **The plan disclosure survived the chooser it lived on.** It now sits on the row that leads to a bank, once per screen. A resumed signup can meet the hub-with-list before ever seeing the empty one, so the disclosure has to be there too.
- **Free signups.** Still offered no bank connection. On an empty hub that leaves one row, so they go straight to the manual form as before. On the hub-with-list they keep the two suggestions whose wording promises no connection ("a mortgage or a loan", "another bank"); the token-only one is dropped.
- **#982/#983** left untouched — `connect-account-inline.tsx` is not modified.

## What the design asks for and this does not do

The mockup shows per-account balances and movement counts ("612 movements · 12 months", "€2,340.11"). The onboarding payload carries neither — balances live in a separate table synced to IndexedDB, and movement counts have no endpoint. Adding them means a new server payload, which is not this PR. The rows carry `Checking · EUR`, as the old list did. Worth picking up if the reward framing matters.

## Instrumentation

Extends #981 rather than redoing it:

- `onboarding_step_viewed` gains `accounts_count` — the only thing that separates the hub's two states in a funnel, since they share a step.
- `onboarding_accounts_hub_route` reports which way out was taken (`connect`, `manual`, `mortgage`, `investment`, `bank`) with the account count. That is the question the step event cannot answer, and the one this screen exists to ask.

## Gates

`pint --dirty` · `bun run lint` · `bun run format` · `bun run dry` (3.62%, down from 3.63%) · `crap --base=origin/main` · `vitest` 876/876 · `pest --exclude-testsuite=Browser` (10 pre-existing 409 failures in `DashboardTest`/`SavingsGoalTest`/`SharedAccountOwnershipTest`, reproduced on the base branch, no PHP source touched here) · onboarding browser suite 31/31.

One browser fix worth flagging: the end-to-end test clicked `Continue` twice after the CSV import. The import step completes itself, so the first click was already landing on the accounts list — one click on the hub's own button now covers both.


victor-falcon:

PR #992, opened 2026-09-14
feat(onboarding): ask first, deliver back — the new step machine and steps 1-5

victor-falcon:
## What

The base of the onboarding redesign: a new step machine, the answers the user
gives us, and the first five screens of the new flow.

The old onboarding explained the product before handing anything over — three
screens of theory in a row (`account-types`, `category-types`, `smart-rules`).
This one asks first and gives back after, and every question the user answers
comes back to them as something.

## The five new steps

| # | Step | What it does |
|---|---|---|
| 1 | `promise` | A claim, not a welcome |
| 2 | `goal` | What they want to change — feeds the target at the end of the flow |
| 3 | `today` | How they track their money now — says what they bring with them |
| 4 | `guess` | What they think they spent last month |
| 5 | `plan` | Their own answers, read back, with what happens next |

**Step 4 is the hinge.** It costs one drag and opens a question only the app can
close; a later step answers it with the real number. It is also the only answer
the flow actually needs — the other two are stored because the question was
asked anyway (product owner's call).

## Persistence

New `users.onboarding_answers` JSON column, written through
`POST /onboarding/answers` (`StoreOnboardingAnswersRequest`, `Rule::in` on both
choice lists). Each question saves as it is answered and the write **merges**,
so quitting on the third one still leaves the first two behind. The spending
guess is stored in the minor units of the user's own currency, like every other
amount in the app.

The column comes back as an `onboardingAnswers` prop, so a reload on `plan`
still reads the answers back instead of degrading to the plain copy.

## What is deleted

`account-types`, `category-types`, `smart-rules` and `customize-categories`,
with their components and their entries in `VALID_STEPS` / `PRIMARY_STEPS`.

`customize-categories` was also a dead end: it was in `VALID_STEPS` but not in
`PRIMARY_STEPS`, so `goNext()` did `indexOf()` → `-1` and never advanced, and it
was not in `BACKABLE_STEPS` either. Anyone who landed there was stuck.

`welcome` goes too — `promise` replaces it as step 1 (a claim, not a welcome),
which is why the progress bar reads 1/11 on it, matching the design.

## The stored resume point — the risk, and how it is handled

The current step lives in `localStorage` under `onboarding-step`, owned by the
user id. The step names change here, so anyone mid-onboarding at deploy time has
a step stored that no longer exists.

No new key version was needed: `readStoredStep()` already validates the stored
value against `VALID_STEPS` the same way a deep link is validated, so a retired
name resolves to nothing and falls back to the first step. The steps that
outlived the rename (`create-account`, `syncing`, …) still resume, so a user who
had already created their accounts is **not** sent back to question one.

Both halves are covered: a parameterised test over all five retired names, and
one asserting a kept step still resumes.

## Progress bar

Drawn over **eleven** — the length of the finished flow — rather than over
`PRIMARY_STEPS.length`. The five steps behind `plan` are still the old ones
until the later PRs replace them; measuring against the list would walk every
percentage back a notch each time one of them lands.

## Instrumentation

`onboarding_step_viewed` (from #981) carries the new step names with no change;
`total_steps` becomes 11 (10 for a free signup, which still skips the AI step).

Added alongside it, not instead of it: **`onboarding_answered`** with `question`
and `answer`, fired from the same place the answer is written. The spending
guess reaches PostHog as a number, so the distribution is queryable from day one.

## Not in scope

Step 6 onwards, billing, the payment gates and the three account-creation paths.
Five later PRs.

## Screens

Captured by `tests/Browser/OnboardingQuestionsScreenshotsTest.php` at 430×932.

| 1. Promise | 2. Goal | 3. Today |
|---|---|---|
| promise | goal | today |

| 4. Guess | 5. Plan |
|---|---|
| guess | plan |

The `plan` screen is the reciprocity working: it carries back the goal picked on
screen 2 and the number set on screen 4, both in the user's own words and
currency.

## Tests

- **`use-onboarding-state.test.tsx`** — 35 tests: the renamed machine, the
  retired-step fallback, the kept-step resume, the answers (local state, the
  merge, the request shape, the analytics event) and the new step counts.
- **`step-choice.test.tsx`** — the shared question screen: selection, and the
  action held shut until something is picked.
- **`step-guess.test.tsx`** — the slider display, minor-unit conversion, and
  resuming from an answer already given.
- **`step-plan.test.tsx`** — the answers read back, and the fallback copy when a
  deep link lands there with nothing answered.
- **`OnboardingAnswersTest.php`** — the endpoint: storing, merging, rejecting an
  option that was never offered, and a non-integer guess.
- **Browser** — the four questions walked end to end (with a DB assertion on
  what was stored), the full end-to-end flow, and the phone-width walk.

One test was deleted: `shows real estate on onboarding account types by
default`, which only asserted the contents of the `account-types` screen that
this PR removes.

## Gates

`pint --dirty` · `bun run lint` · `bun run format` · `bun run dry` (3.67%, under
the 5.4% threshold — `StepChoice` is shared by `goal` and `today` precisely so
the two do not land as a clone) · `php artisan crap --base=origin/main`
(5 methods, none over 10) · `pest --exclude-testsuite=Browser` — 3279/3290, with
10 pre-existing failures in `DashboardTest`, `SavingsGoalTest` and
`SharedAccountOwnershipTest` (Inertia 409 from the local asset manifest).
Verified identical on `origin/main`: same 10, same assertions, nothing to do
with this change.

Translations added for ES and FR.


victor-falcon:

PR #993, opened 2026-09-14
feat(onboarding): the bank path, from the picker to both ways back

victor-falcon:
The "connect a bank" route that hangs off the step 6 hub, end to end — including the two returns that had no design at all.

Stacked on `onboarding-accounts-hub` (#990). Review that one first.

## The eight screens

| | |
|---|---|
| **Connect** | `Where do you bank?` — the bank search, with the country as a control beside it |
| **Country** | the full list, reached from that control or when the guess finds nothing |
| **Handoff** | `You're about to log in at BBVA` — the last screen before the bank |
| **Reading** | the wait, with counters that move instead of a spinner |
| **Taking its time** | the stuck state, and a way out of it |
| **Which accounts** | which of the accounts the bank returned actually come in |
| **Connect failed** | the return from a bank that refused |
| **Other device** | the return that landed in a browser with no session |

## Why the country stopped being a step

`EnableBankingProvider::getInstitutions(string $countryCode)` calls `/aspsps` with `country`
required — there is no global listing — and `startAuthorization(string $aspspName, string $countryCode, …)`
is keyed by the `(name, country)` pair. Santander exists in ES, PT, PL, DE and GB, so without a country
the authorization goes to the wrong ASPSP.

It is part of a bank's identity, then, not a filter. But it is also guessable: `useGuessedCountry()`
reads the region off the locale the user already formats their money in, and the flow opens on that
country's banks with the full list one tap away. A region we do not connect in falls back to the
country list as the first screen, which is the only case it is still a step.

The search placeholder counts the list — "Search 214 banks" — which says you are searching inside one
country without having to explain it.

## The beta flag is the reliability signal

Already surfaced as a pill on the bank row and a notice on the handoff (both pre-existing), and both
survive the redesign. Worth restating why, from `EnableBankingProvider::getInstitutions`: *"over a week
of calls, its beta connectors failed eleven times as often as the stable ones, and they are left out of
the provider's public status page, so nobody reports their outages either."* It is information the user
wants **before** choosing, not after getting stuck.

## Which accounts — a behaviour change

Onboarding used to take every account a bank returned, without asking
(`AuthorizationController::completeFirstConnection` → `createAccountsFromPending`). That is how a joint
account or a card nobody uses ends up in a first picture, and it is hard to undo later.

Now: **one** account is not a decision, so it is still created straight away and synced. **More than
one** leaves the connection `AwaitingMapping`, and the hub renders the chooser. It posts to the
existing `open-banking.map-accounts.store` endpoint, so `skip` means there exactly what it already
means in settings — nothing is created, and it can be brought in later.

`AccountMappingController::show` no longer auto-creates during onboarding either, or it would defeat
the chooser for anyone who reached that URL directly. It redirects back to the hub, which asks.

## Reading, with real numbers

`OnboardingController::syncStatus` now also returns what has actually landed: transactions, distinct
counterparties, accounts, and the months the import spans. Deliberately **not** a progress bar — a bank
never tells us how much it is about to hand over, so a bar would have to invent its denominator. A
number that only goes up promises nothing it cannot keep.

Counterparty names are used for "places you spent" because `description` is encrypted at rest and
cannot be counted server-side.

The stall now splits in two, which it did not before: a bank that is **slow** gets "nothing is broken"
and the choice to carry on with what is in or keep waiting; a sync that actually **failed** keeps the
old copy. The previous screen said the same thing to both, and for the slow case it was not true.

## The two returns that had no design

**Connect failed.** `handleAuthorizationError` deletes the unfinished connection, so by the time the
user is back there is no row left that knows which bank refused. The bank and its country now travel
back on the URL, which is what lets the screen name it and makes "Try BBVA again" one tap instead of a
fresh search. The first two rows exist because an interrupted bank login looks, from the user's side,
like something half-done was left behind.

**Other device.** `finishRedirect` renders `/open-banking/connection-complete` when the callback
arrives without a session — the iOS PWA case, where the bank hands the return to Safari. That page was
plain text with **no button at all**. It now names the bank, says the connection worked and is already
downloading, and has a way back.

## Screenshots

| **Connect** | **Country** |
| --- | --- |
|  |  |

| **Handoff** | **Reading** |
| --- | --- |
|  |  |

| **Taking its time** | **Which accounts** |
| --- | --- |
|  |  |

| **Connect failed** | **Other device** |
| --- | --- |
|  |  |

Captured by `tests/Browser/OnboardingBankPathScreenshotsTest.php` at 430×932, same convention as #981 and #990.

## Deliberately not in this PR

- **API-key connectors their own section.** The Connect mock groups Indexa, Coinbase and the rest under
  "Brokers and exchanges". They already appear in the same picker today (`CONNECT_PROVIDERS` is merged
  into the institution list) and still do; giving them their own section is PR F. Their confirm step
  keeps its own copy rather than the handoff's, because none of the "you log in at your bank" text is
  true of an API key.
- **Per-country bank counts.** The Country mock shows "214 banks" against every country. Each one costs
  a `/aspsps` call, so 19 provider calls to draw one list. Dropped; the count still appears on the
  Connect screen, where we have already fetched it.
- **One extra button.** The Connect mock's footer has only "My bank isn't in the list". There is also a
  ghost "Back", because the onboarding header's arrow is not available here — `BACKABLE_STEPS` excludes
  `create-account`, and everything hanging off the hub carries its own back control (the manual form
  already does).
- Billing gates and the paywall are PR L.

## Tests

- `tests/Browser/OnboardingBankPathScreenshotsTest.php` — the eight screens, captured at 430×932.
- `BankConnectionFlowTest` updated: the onboarding walk now goes list → handoff → chooser, and asserts
  nothing is created until the user picks.
- New coverage for the multi-account callback, the `pendingMapping` prop, the sync counters, the
  chooser's skip payload, and the failure return's URL handling.
- `usableCurrency()` moved from the settings mapping page into `lib/banking-connections.ts` so the
  chooser reuses it rather than growing a second copy.


victor-falcon:

PR #995, opened 2026-09-14
feat(onboarding): the broker path, for the accounts a bank never returns

victor-falcon:
Stacked on `onboarding-bank-path` (PR E). Chain: C → D → E → **F**.

## What

Whisper has six API-key connectors — Indexa Capital, Coinbase, Binance, Bitpanda, Wise and Interactive Brokers — and the onboarding had no idea they existed. They are not a footnote: PSD2 returns current accounts and cards, so a pension plan or a broker arrives through one of these or it does not arrive at all.

A broker is not a bank with a different form. No redirect, no bank login page, and none of the handoff's reassurance is true of an API key. So they stop being rows in the bank list and get a path of their own — driven by the `CONNECT_PROVIDERS` registry that already describes every one of them (endpoint, credential fields, labels, help link). Nothing about a provider is written twice.

## Three ways in, one screen out

1. **The hub** — "A pension or a broker" now opens the broker list instead of the bank picker. That row sits under *usually missed* rather than next to the manual form, and the distinction was the point: a broker is connected, not typed in.
2. **The bank picker** — keeps them, as a **Brokers and exchanges** section under the banks instead of mixed into them, folded after the first two. They were searchable when they were rows in that list, so the section is filtered by the same search box — typing "Binance" still finds it.
3. Either way the user lands on the same credential screen.

| Brokers under the banks | The list the hub opens |
| --- | --- |
|  |  |

| The credential screen | The one provider that does bring movements |
| --- | --- |
|  |  |

## Holdings, not movements — except for Wise

The screen says the thing the flow could not say before: brokers report holdings, not purchases, so the account moves as one number and **no transactions are coming**. Promising a statement that never arrives is the failure mode this copy exists to prevent.

It is also the one place the registry needed a new field. Wise has a `WiseTransactionSyncService` and *does* hand over movements, so it gets the sentence that is true of it instead. `sendsTransactions` marks that single entry; the silence of the other five is the statement about the group. Same flag drives the row subtitle (`API Token · holdings, not movements` vs `Personal API Token · balance and movements`), so what a provider will and will not deliver is said before the user spends a screen on it.

The key is read-only and the screen says so, next to where each provider hides it (`help.link`) and Coinbase's IP-allowlist condition, which its own form gets wrong by default.

## Deliberate calls worth a second opinion

- **Copy deviates from the design in one sentence.** `ConnectBroker.dc.html` says "Brokers and pension funds don't use open banking". Wise is neither, and it is in this list, so the screen says "These don't use open banking" instead. Everything else is the design's wording.
- **Coinbase moves up the registry**, so the picker folds to Indexa + Coinbase as the design has it rather than Indexa + Binance. The settings dialog sorts alphabetically, so nothing changes there.
- **`aspsp_country`** travels with the connection for the three providers whose endpoint requires it. It is a label on a global exchange rather than part of its identity — `WiseController` hardcodes `'GB'` server-side for the same reason — so the bank picker hands over the country the user already chose, and the hub, which has no picker in front of it, falls back to `useGuessedCountry()`.

## Reuse, not new plumbing

- `useConnectFlow` grows one option, `separateProviders`: the settings dialog keeps showing them inline, only the onboarding splits them out.
- The inline flow's confirm step loses its provider branch entirely — it is now only ever the bank handoff.
- Both flows POST through one shared `postConnectRequest`, which is where the duplicated fetch was heading.
- `ProviderCredentialInput` comes out of `ProviderCredentialFields` so the onboarding can pair the same control with `StepField` instead of the dialog's `Label`. (It drops a stray `mt-1` from the settings dialog's inputs: 4px, spacing the container already owns.)
- `StepCallout` and `StepEmphasis` come out of the hub rather than being copied into a fourth screen.

**No backend change.** The endpoints have worked outside the wizard for months; this PR only gives them a door. Billing and the gates stay exactly where PR E left them (PR L).

## Tests

- **Vitest** — `connect-broker-inline.test.tsx` (11): the list, what each provider promises, the credentials the registry names, the no-transactions warning and its Wise exception, the disabled-until-complete button, the POST body incl. the country, the analytics event, a refused key keeping the user on the form with what the provider said, and both back behaviours.
- **Vitest** — `connect-account-inline.test.tsx`: the section exists and is folded, a folded broker is still reachable by search, and a broker row opens the broker form rather than the bank handoff.
- **Vitest** — `step-accounts-hub.test.tsx`: the pension row goes to the brokers, not to the manual form.
- **Browser** — `tests/Browser/OnboardingBrokerPathScreenshotsTest.php` (4), which produced the screenshots above: `broker-section`, `broker-list`, `broker-connect`, `broker-connect-wise`.
- ES + FR translations for every new string.

Gates green: `pint --dirty`, `bun run lint`, `bun run format`, `bun run dry`, `php artisan crap --base=origin/main --no-coverage`, `pest --exclude-testsuite=Browser` (3355 passed).

One unrelated line of prettier drift in `step-import-balances.test.tsx` came along with `bun run format`.


victor-falcon:

PR #996, opened 2026-09-14
feat(onboarding): the file import as screens, not a drawer on top of the flow

victor-falcon:
Stacked on `onboarding-broker-path` (PR F). Chain: C → D → E → F → **G**.

## What

The import inside the onboarding was not a screen, it was a modal: `ImportTransactionsDrawer`, 999 lines of the app's generic UI — its own `DrawerHeader`, its Title Case titles and its dashed drop box — opened on top of the wizard. It was the last place the old design got in.

Its five steps (`SelectAccount → UploadFile → MapColumns → Preview → Importing`) become seven screens of the flow, built from `StepScreen`, `StepList`, `StepRow` and `StepButton` like every other step.

**The drawer is not deleted.** The transactions page still depends on it, and it is untouched from the outside.

## The seven screens

| Which account is this file from? | Bring in your history |
| --- | --- |
|  |  |

| Did we read it right? | The movements, before anything is saved |
| --- | --- |
|  |  |

| Bringing them in | Some in, some we couldn't read |
| --- | --- |
|  |  |

| The PDF every bank offers first |
| --- |
|  |

## "Which account is this file from?" becomes a real question

The drawer skipped that step with `autoSelectSingleAccount`, which was true while the onboarding created exactly one account. The hub from PR D allows several, so the assumption is gone: the screen appears as soon as more than one account can take a file, preselecting the one just created.

Two kinds of account are on the list to be ruled out rather than picked:

- **Connected** — the bank is already sending its movements. A file on top of one is exactly how someone ends up with a year of their history twice, and the callout says so.
- **Balance-only** (pension, broker, mortgage, property) — there is nowhere to put rows.

`autoSelectSingleAccount` had no other caller once this step stopped passing it, so the flag, the auto-select effect and the back button it used to hide are gone from `ImportStepAccount` and the drawer. One account in the drawer now means one click, not zero.

## One import engine, not two

Copying the import loop into the wizard would have put the partial-import fix from #988 in two places and tripped `bun run dry`. So the parts both need move to `lib/transaction-import.ts`:

- `parseImportFile()` — read the file, guess the columns, sample three cells from the middle of each one.
- `applyStoredImportConfig()` — overlay the layout this account was last imported with, ignoring one that names columns this file has not got.
- `importTransactions()` — create the rows in batches, apply the automation rules, write the balances, and report which rows made it.

The drawer now calls all three and keeps its own toasts and state; it is **354 lines shorter** and its behaviour is unchanged, including the two #988 tests, which still pass untouched apart from picking the account by hand.

## Two lines of the design's copy were not true, and now say what is

- `UploadFile.dc.html` promises "nothing is **uploaded** until you've seen the preview". Parsing is genuinely client-side (SheetJS + `FileReader`), so *read in your browser* holds — but `checkDuplicates` POSTs each row's date, amount and description before the preview renders. The screen says we check what you already have, and that nothing is **saved** until you say yes.
- `Importing.dc.html` says leaving "cancels the import and keeps nothing". Rows are created one at a time and stay created, so it says the import stops where it got to and what is already in stays in. Cancelling for real would mean rolling back rows the user asked for; not worth it for a screen that lasts seconds.

## Smaller calls worth a look

- **The results screen counts rows the mapping dropped**, not just rows the server refused. `buildMappingReport()` already reports them with their real line numbers, so "2 rows · No date · Rows 314–315" points at the totals line at the bottom of the file — which is what the design's copy is about, and something the drawer never told anyone.
- **"Connect the bank instead"** returns to the hub, where "Connect a bank" is the first row, rather than opening the bank flow directly — that would mean a new prop threading through the hub from PR D. One extra click, no new coupling. It is hidden on the free plan, which has no connections to offer.
- **A 10 MB limit** now exists because the screen promises one. Parsing happens in the browser, so a file big enough to hang the tab is turned away before it is read.
- **The header's back arrow still leaves the step for the hub.** Moving backwards *inside* the flow is on the screens themselves ("Use a different file", "Choose another file", "Fix the file and retry"), the way the bank and broker flows carry their own back buttons.
- **`:count months`, not "six months"** — spelling numbers out does not survive translation.

## The exit from #989

That fix is in `main` but not in this stack's base, so it is not in the diff. The behaviour is here anyway, and unconditionally: **"I don't have one yet"** sits under "Choose a file" on the upload screen, rather than appearing only after the drawer had been opened and closed once. `step-import-transactions.tsx` is rewritten wholesale, so resolving that conflict at merge time means taking this side.

## Tests

- **Vitest** — `step-import-transactions.test.tsx` (10): asking or not asking which account, a connected account not counting as one, a PDF going to the error screen without being parsed, the preview writing nothing, the exit, the clean import moving on, a partial one staying put, the dropped rows being reported, and a retry re-checking duplicates so nothing is created twice.
- **Vitest** — `step-import-pick-account.test.tsx` (7) and `step-import-partial.test.tsx` (8): eligibility, and how row numbers read (`Rows 44–46`, `Rows 112, 140, 203`, `Rows 1, 3, 5 and 2 more`).
- **Browser** — `OnboardingFileImportScreenshotsTest.php` (6) produced the screenshots above from a real 314-row CSV: six months of movements plus the two-line totals block at the bottom. The progress shot waits for a batch boundary (`60 of 312`) rather than racing the import.
- **Browser** — `OnboardingFlowTest.php`'s full walkthrough now goes through the screens instead of the drawer.
- ES + FR translations for all 73 new strings, including the five parser fault reasons (`No date`, `Amount can't be read`…) which reach the user for the first time.

**One test deleted:** `import-step-account.test.tsx`'s "auto-selects a single account when enabled for onboarding" covered `autoSelectSingleAccount`, which this PR removes. The other test in that file, which asserts a single account is *not* auto-selected, stays and is now the whole behaviour.

Gates green: `pint --dirty`, `bun run lint`, `bun run format`, `bun run dry`, `php artisan crap --base=origin/main --no-coverage`, `bunx vitest run` (924 passed), `pest --exclude-testsuite=Browser` (3345 passed).

**Pre-existing failures:** 7 in `tests/Feature/DashboardTest.php` (`top_labels`/`top_categories`, HTTP 409). They fail identically on `onboarding-broker-path` at 7074930e with none of this applied.


victor-falcon:

PR #997, opened 2026-09-14
The reveal: answering the guess the flow opened with

victor-falcon:
Stacked on `onboarding-file-import` (PR G). Chain: C → D → E → F → G → **H**.

Step 7, and the reason the flow opens with questions instead of a bank picker.
Step 4 made the user commit to a number; this is where it is answered.

| The reveal | No spending to reveal |
| --- | --- |
|  |  |

## The trap of this screen

**Nothing is categorized yet.** Step 8 writes *rules*, and the batch behind it
runs afterwards — so a breakdown by category is not available, however well it
would read. Merchants are: the name is already on the raw row and needs no model
at all. It also reads harder ("€312 at Mercadona" beats "Groceries €312") and it
sets up the next step, which is exactly those 147 merchants becoming categories.

The closing line is the one observation the raw rows can prove on their own —
*nine merchants charged you the same amount in each of the last three months* —
which is a group-by over merchant and amount, not a model.

## The second variant

`RevealAssets`, for someone who arrived with a pension, a mortgage and a broker
and no current account. They have nothing to be told about their spending, so
they are shown what they are worth and offered the half they are missing.

The condition is deliberately not "does this user have a bank" but "does any
month hold enough movements for the number to mean anything" — five, and the
month need not be last month. A file exported in March is still a real import,
and answering it with an empty month would be worse than answering it a month
late; when the revealed month is not last month the screen names it instead.

## Three cases the design does not show, and the copy survives

- Someone who spent **less** than they feared is not told to account for it.
- A gap too small to name gets the yearly rate instead of a comparison (a €3 gap
  is not "€36 a year you can't account for").
- The closing sentence only points at repeat charges when it actually named some.

Plus: no guess at all (a resumed run), no nameable merchants, and a request that
never comes back — which moves the user on rather than holding them in front of
an apology.

## The numbers

`OnboardingRevealService` loads a year of outgoings rather than aggregating in
SQL, so every amount goes through the currency conversion and the ownership
weighting that every other total in the app does — the same shape
`CashflowSummaryService` already uses. Balance-tracking accounts are left out of
the spending figure: a pension's rows are adjustments, not purchases, and they
would read as a €2,000 spending month nobody had.

It sits behind its own request rather than a page prop: it is several queries,
and the other ten steps would pay for them on every render without ever showing
them.

The merchant count behind the call to action mirrors
`RuleSuggestionAggregator::groupsFor()` — same filter, same
counterparty/description fallback — so the button does not promise a pile step 8
does not work on.

Two of `NetWorthCalculator`'s helpers become public (`countsOn`,
`contributionOf`): the assets variant lists the accounts behind the figure it
prints, and a list built on its own rules would sooner or later disagree with
the sum. `StepEmphasis` is generalized into `StepFilled` (several values, each
with its own emphasis) and keeps its old signature on top of it.

## Tests

- `tests/Feature/Onboarding/OnboardingRevealTest.php` — the month chosen, the
  fallback to an older one, balance-tracking accounts excluded, the recurring
  count, the assets variant, the description fallback, and guests.
- `resources/js/components/onboarding/step-reveal.test.tsx` — the three copy
  branches and the two ways the screen steps aside.
- `resources/js/hooks/use-onboarding-state.test.tsx` — the reveal lands between
  the sync and the AI step.
- `tests/Browser/OnboardingRevealScreenshotsTest.php` — the two screenshots above.

## Out of scope

Step 8's states and Teach are PR I. The goal and the close are PR J.


victor-falcon:

PR #998, opened 2026-09-14
feat(onboarding): the six states of step 8, and Teach

victor-falcon:
Stacked on `onboarding-reveal` (PR H). Chain: C → D → E → F → G → H → **I**.

Step 8 branched into seven render states and the flow treated them as one. Each
one is a screen now, and the step that follows it — Teach — gets the gate it
was supposed to have.

| Generating | Review | Not eligible |
| --- | --- | --- |
|  |  |  |

| Failed | Empty | Re-consent |
| --- | --- | --- |
|  |  |  |

| Teach us (step 9) |
| --- |
|  |

## Two things the copy said that were not true

**The time.** `This can take up to two minutes.` The step usually beats that by a
hundredfold and occasionally misses it. It now says up to a couple of minutes,
and that leaving the screen does not stop it — which is what a queued job
actually does.

**What the step produces.** It writes *rules*, not categories. Nothing is sorted
while it runs, so the empty state no longer claims 847 movements got a category
(the design said that; it is false at this point in the flow) and the review
screen counts rules rather than transactions.

## Counters instead of a spinner

The generating screen shows what the run has reached, not a timer dressed up as
progress. `merchants_considered` is written to the run as soon as the grouping
is done — **before** the model is called, which is the slow half — so the client
polling `show` has a real number while it waits. `suggestions_count` was already
there. One column, one migration.

The cycling status messages (`Analysing your transactions…`, five of them on a
3.5s timer) and the three skeleton cards go with it: they described work nobody
was doing.

## The consent screen knows which question it is asking

Someone who revoked consent, or consented to an older `consent_version`, is owed
the reason they are being asked twice, so `previously_consented` comes down with
the state and picks the wording. A first-timer is not told they switched
anything off. The upgrade notice is untouched — the AI pay gate is PR L.

`AiNotEligible` is the one state with somewhere else to go, so the step takes an
`onAddAccount` and hands the user back to the accounts hub, the same way the
reveal does.

## Teach: the way out that let you through with nothing

`canContinue` was:

```js
const canContinue = isComplete || categorizedCount >= minimumRequired || totalAvailable === 0;
```

`isComplete` is "the index ran off the end of the queue" and `handleSkip` only
moved the index — so skipping every movement got you out of the step with zero
categorized. `isComplete` is gone: the minimum has to be met for real. It is
`min(5, total)`, so it lowers itself when there is little to do.

That alone would trap a user who skips everything in front of an empty screen
with the gate still shut, so a skip now sends the movement to the back of the
queue rather than past the end of it (`recycleSkipped`, onboarding only — the
transactions page still finishes when the queue runs out). The footer can then
promise what it promises: *skip it and we'll bring you another*. Categories being
empty also opens the gate — a user with nothing to file under could never meet
it, and that is not their fault.

The step also loses its preamble screen ("See where you spend", "Build better
budgets", …). After step 8 the user has been told.

## The dispatch from all four exits

PR #985 moved `CategorizeOnboardingTransactionsJob` to the close of step 8 and
hung it off all four exits. They all still leave through `leaveStep()`, and
`step-ai-suggestions-exits.test.tsx` fails if any of them stops doing so.

## Three sentences in the designs that promise a screen we do not have

`AiNotEligible`, `AiEmpty` and `AiFailed` offer to bring the suggestions back
from Settings → Automation rules. There is no such screen: this step is the only
caller of the endpoint. They say what is true instead — any transaction can
become a rule from the app itself.

## Tests

- `resources/js/components/onboarding/step-ai-suggestions-states.test.tsx` — the six screens, including the two consent wordings and the review screen with everything unticked.
- `resources/js/components/onboarding/step-ai-suggestions.test.tsx` — rewritten: the counters and the honest duration line, in place of the three tests that asserted the cycling messages (the copy they assert no longer exists).
- `resources/js/components/onboarding/step-categorize-transactions.test.tsx` — new: the gate opens on five filed, stays shut on six skipped, and a skip brings another movement round.
- `tests/Feature/Ai/GenerateRuleSuggestionsTest.php` — the counts are on the row before the generator is called, and survive an empty run.
- `tests/Feature/Ai/RuleSuggestionControllerTest.php` — `previously_consented`, and a run in flight reporting its merchants.
- `tests/Browser/OnboardingAiStatesScreenshotsTest.php` — the seven screenshots above.
- `tests/Browser/OnboardingFlowTest.php` — the walk-through now files five movements instead of skipping five, and clicks through the reveal, which PR H put between the hub and this step.

`lang/es.json` and `lang/fr.json` carry the 64 new strings.

## Not in here

The AI pay gate (`GateAi`) is PR L; the goal and the close are PR J.

`AiReview` prints "84 movements · €5,210" per rule and `Teach` prints "Seen 31
times in the last year". The count is real (the server counts the matches); the
money and the per-merchant history are not available client-side — the merchant
grouping lives in `RuleSuggestionAggregator`, behind the run. Wiring that into
the review and the categorizer is a change of its own, not a copy tweak.

Two tests in `OnboardingFlowTest` are red on `onboarding-reveal` before this
branch and stay red: `it hides the connected plan warning…` looks for `Connect
Your Bank`, a screen the hub PR replaced, and the bank-authorization-failure one
does not get back to the hub. Neither touches this step; they belong to the PRs
that renamed those screens.


victor-falcon:

PR #999, opened 2026-09-15
feat(onboarding): the first target, and a close that only claims what exists

victor-falcon:
Stacked on `onboarding-ai-states` (PR I). Chain: C → D → E → F → G → H → I → **J**.
Last of the redesign: it closes the flow.

| 10 · First target | 11 · Ready |
| --- | --- |
|  |  |

## The decision that governs this PR

The target is **optional**. The product's non-negotiable was accounts created,
movements imported and categorized; a target never was. So it can be declined
without friction — and it sits at the end of a long flow, right after the user
paid, which is the worst possible place for a wall.

**So step 11 is built from the user's own data, not from a template with
conditionals bolted on.** Each line is asked whether it is true and dropped when
it is not. No target, no target line. No bank, nothing about accounts keeping
themselves current. A closing screen that oversells is worse than a sober one,
and the dashboard behind it would contradict it within seconds.

| Bank + AI + target | File only, target declined | One balance, nothing else |
| --- | --- | --- |
|  |  |  |

The third user has one true line, and the screen says so rather than padding it
out — but an empty half-screen is not "still worth reading", so what fills the
rest is the one thing that would fill the list: connect a bank, or bring a file.

## Step 10 is the crossing of two answers the user already gave

`GET /onboarding/summary` reads what they actually have. The target opens on a
tenth of the month **step 7 read back to them**, not on a figure we invented,
and the sentence under the title sets it against the guess from step 4.

The goal from step 2 picks the second half of that sentence. Step 2 promised the
answer would decide "what your first target will be", so a target that reads
identically for someone paying down a debt and someone saving for a year off
would have made that a lie.

The comparison in the grey block is only drawn when there are repeat charges to
draw it against — `recurring_amount` is new on the reveal service, computed from
the raw rows the same way `recurring_count` already was. No repeat charges, no
equivalence: an invented one is exactly what this screen exists to avoid.

## What "Set my target" actually creates

A **monthly catch-all budget**, limit = the month they really spent minus what
they chose to keep back. Catch-all so every movement counts towards it without
asking someone on their way out of onboarding to assign categories to a budget.
That is what makes the target a real thing the dashboard can show, and what lets
"we'll tell you the day you drift off it" be true — the toggle writes
`notify_on_close_to_limit` / `notify_on_over_limit` directly rather than leaving
them to the account default, so the row that says it is the row that decides it.

A `SavingsGoal` was the other candidate and was wrong: with no linked label its
`savedAmountInCents()` is 0 forever, so the dashboard would show a goal stuck at
0% — the exact kind of lie this PR is about.

The target is remembered in `onboarding_answers` **only after the budget
exists**, because step 11 reads it back as proof the target is real. A user who
already has a catch-all keeps it: a reload must not leave two budgets claiming
the same movements.

## The eleven steps stop being a hardcoded number

`TOTAL_PRIMARY_STEPS = 11` was a placeholder for a list that had not been
finished yet. The list is finished, so the bar reads the list.

That also puts `syncing` where the design always had it: sharing the sixth
position with the accounts hub rather than taking one of its own. Waiting on a
bank's first sync is part of connecting it — the bar was moving for something
the user did not do, and everything after it was off by one.

## Two shared reads instead of two copies

`syncProgress()` left the controller: the sync step and the closing screen ask
the same question of the ledger (how much, how far back), so it is one
`OnboardingSummaryService::ledger()` now. The controller is 50 lines lighter.

## `complete()` is untouched

It still stamps `onboarded_at` and redirects. The AI batch stays where #985 put
it — the close of step 8 — and has not been moved back.

## Tests

- `tests/Feature/Onboarding/OnboardingTargetTest.php` — the summary, the budget
  it creates, the limit it sets, the idempotent second run, the 422 for a user
  with no month to build on, and the target read back.
- `step-target.test.tsx` (15) — the opening figure, the stepper's floor and
  ceiling, both exits, the failure that keeps the button alive, the goal framing,
  and the two ways it steps aside.
- `step-complete.test.tsx` (14) — the three users above, plus the summary that
  fails to load and still lets the user out.
- `OnboardingCloseScreenshotsTest.php` — the four screenshots above.

## Not fixed here, but noticed

Two browser tests in `OnboardingFlowTest` are red on this branch **and on its
base**, both left behind by earlier PRs in the stack rather than by this one:

- `it hides the connected plan warning…` expects `Connect Your Bank`, copy that
  PR G replaced with the country picker.
- `it returns to the accounts step when bank authorization fails…` expects the
  hub; PR G made a failed authorization land on `ConnectFailed`, which is the
  correct new behaviour.

Left alone so this diff stays about step 10 and 11. They are in the `Browser`
suite, which is not part of the merge gate.


victor-falcon:

PR #1000, opened 2026-09-15
feat(onboarding): the benefits-first onboarding, and the gates that ask before the money is spent

victor-falcon:
The onboarding redesign, whole, with the two rounds of browser QA it has been through folded in. It was reviewed as a chain of nine stacked PRs (C → D → E → F → G → H → I → J → L), all closed in favour of this one; #1004, which carried the first QA round's ten fixes, is merged into it as well. **Review it commit by commit** — each one is self-contained and its own message carries the argument.

## The nine the redesign is made of

| | Commit | What it does |
| --- | --- | --- |
| **C** | `498a74ee` | Ask first, deliver back — the new step machine and steps 1‑5. The flow opens on a promise and four questions instead of a form. |
| **D** | `13464507` | Step 6 becomes an accounts hub you come back to, not a fork you pass through once. Banks, brokers, files and manual accounts all mix. |
| **E** | `ecb6eb2d` | The bank path, from the picker to both ways back — country screen, hand‑off screen, mapping, first sync. |
| **F** | `7074930e` | The broker path, for the accounts a bank never returns — Indexa, Coinbase, Binance, Bitpanda, Wise, Interactive Brokers, by token. |
| **G** | `78e133d4` | The file import as screens, not a drawer on top of the flow — upload, column check, preview, progress, partial. |
| **H** | `00e2b005` | The reveal, which is why the flow opens with questions: what they actually spent, against what they guessed. |
| **I** | `7a2a1f9b` | The six states of step 8, and Teach — consent, working, rules, nothing, failure, already‑on. |
| **J** | `6a03f741` | The first target, and a close that only claims what exists. |
| **L** | `428cd5c8` | Pay now, and the gates that ask before the money is spent. |

## The payment rule, whole

A bank connection and an AI run both cost us money per user from the moment they are switched on, and both were handed out during onboarding against a plan chosen at the end. This turns that order around: the plan is bought where the feature is reached for, and the feature is switched on after.

1. **Charged in full at signup.** `trial_days` is 0 on both plans. The way back out is the three‑day self‑service refund #986 restored.
2. **No bank and no AI without a subscription.** The wizard says so at the gate; `HandlesSubscriptionGate` says so where it cannot be talked past, onboarding no longer excepted. The gate comes up in front of the broker picker too, because `OpenBankingConnectController` gates those identically — without it Indexa and Coinbase would hand back a 402 mid‑wizard.
3. **The AI consent travels with the plan**, recorded at the checkout the gate started. That is where the user gave it: they read the disclosure row and pressed the button.
4. **The free plan stays** for whoever connects no bank and switches on no AI.
5. **The close**: straight through for a subscriber, soft paywall with a free exit for everyone else.

### The disclosure row is not decoration

Grouping the consent with the purchase is only honest if the purchase screen says, where nobody can miss it, what leaves the account — merchant name and amount — and what never does — balance, name, account number — and that one tap in Settings turns it off. Both gates carry that row. A consent nobody read is not a consent, and the site promises *"your data is never shared with third parties"*.

### `--success`

The money‑back row is the first colour with chroma in the interface proper, and it is deliberate the same way `--destructive` is. It lands in `app.css` as `--success` / `--success-fill` / `--success-border`, measured at 5.7:1 on its own fill, with dark values alongside.

## The gates

| GateBank | GateAi | PaywallSoft |
| --- | --- | --- |
|  |  |  |
| Before the bank picker. | Step 8, for whoever came the manual way — anyone who connected a bank paid at the other gate. | The close without a plan, and the only close a new signup without one can reach. |

| PaywallFormer | FreePlanConfirm | GateBank, dark |
| --- | --- | --- |
|  |  |  |
| Outside the wizard: paid, then stopped. | Was a dialog; now a screen, because it can name the banks it is about to cut. | The money-back row on the other ground. |

All six come out of `tests/Browser/PayNowGatesScreenshotsTest.php`, which reads the copy rather than counting elements — what a checkout screen says about the charge is the thing under review. The screens of the eight PRs below it are in the same style, one screenshot test per step.

## The fixes on top of them

Two browser QA rounds, each driving the flow end to end and checking the database after. The first produced #1004's ten (`dd662a31`…`9d31b396`), all re-verified here. The second walked paths the first had not — a bank and a hand-added account in the same run, the broker gate, a phone with a dropped request — and produced six more.

| Commit | What it fixes |
| --- | --- |
| `e71da2b5` | **A hand-added account lost its import screens once a bank was connected.** The step asked "are there any accounts yet", true from the moment the bank's seven landed in the props while the one just typed in was still missing. None of the bank's can take a file, by design, so the step read that as nothing to do and handed itself back to the hub: account made, history never asked for. |
| `82934d53` | **The broker gate sent you back to the bank picker.** The source was picked by "is this the AI gate or not", so a broker checkout parked the bank return and came back to a country list — asked of somebody who had just paid to connect Indexa. Broker conversions were filed as bank ones for the same reason. |
| `20a065ab` | **A toast landed on the step's own button.** At 390px the network toast sat at 755–828 over a "Let's go" button at 740–792, and the click was refused as intercepted — on the screen where a dropped request most makes a user want to press it again. |
| `3883a931` | The hub's browser tab still read "Onboarding - Create Account", through the hub, the manual form and all five import screens. |
| `0bcb98fd` | One browser assertion the manual-form rewrite missed. It fails `browser-tests`, a required check. |
| `da428cd4` | Two reveal suites declared the same global `revealUser()`. CI never saw it because it runs Browser in its own job, but `pest --filter=` — how CLAUDE.md says to run one test — died on a fatal before anything executed. |
| `ae50e575` | `tests/Browser/live/connect-bank.mjs` still drove the pre-redesign wizard and asserted copy this PR replaced, so the one scenario covering a bank connected *inside* the wizard had been failing since the redesign landed. Four of four now. |
| `358b7624` | **Pay-now goes out behind a switch, off.** `SUBSCRIPTION_PAY_NOW` decides the plans' `trial_days` default and nothing else, because `trial_days` is what `ExperimentOffer`, the checkout terms and the refund button already read. Off — what ships — the plans keep the trial `main` sells today. Two things had to stop lying once the trial could come back: the money-back row promised three days to change your mind next to "free for 15 days", and the button said "€53.94 today" over terms saying nothing would be charged today. |
| `cef7ef56` | **The reveal stopped calling its number spending.** Nothing is categorized when it runs, so the figure carries the user's own transfers. On production, internal movements are 34% of what leaves an account; 52% of users would have the number inflated by over 30%, 29% by over half. It read the gap as a verdict — "you were out by €92 — that's €1,105 a year you can't account for" — to people whose €92 was their own savings transfer. |
| `638249a1` | **A mortgage was asked for its balance twice**, once in the form and again on the screen behind it. |
| `4777b3cc` | The AI step drafts rules from 40 movements instead of 50: a quarter of a year of a file lands around 40 and was told there was nothing to learn from. |
| `1cae63df` | **A spreadsheet date read as `46243` on the column check.** The reader keeps Excel dates as serials on purpose — converting on the way in would coerce a CSV's text dates — which is right for the parser and unreadable on the one screen asking the user to confirm the column. |

### What was verified, in a browser, against this branch

- **The whole flow end to end**, twice over: the four questions, the hub, the gate, a real Stripe test checkout, a real Enable Banking sandbox connection through SCA, the mapping, a 312-movement file import, the reveal, the categorizer, the target and the close — with `agent:db` run after each step.
- **The reveal answers a month that has finished.** Eleven months plus six movements dated this month reveals **Last month · €1,292** with the verdict intact; September is not offered. This is #1004's first finding, re-measured.
- **Real checkout**: €53.94 charged today on the high tier, `trial_ends_at: null`, subscription `active` through the webhook, the AI consent recorded with it, and "Pick up where you left off" returning to the picker the gate stood in front of.
- **The bank path against the sandbox**: picker → country → hand-off → BBVA login → SCA → callback → mapping → 7 accounts → `active` → 221 movements synced. `connect-bank.mjs` passes all four scenarios.
- **The free path**: the AI gate, its free exit, the categorizer's guard, the target, the close, the soft paywall and "Carry on free" onto the dashboard, with no subscription row.
- **At 390px**: nothing scrolls sideways on any screen (`scrollWidth === innerWidth === 390` everywhere), and the last line clears the pinned footer by 48px on the bank gate, 48px on the AI gate, 66px on the populated hub and 48px on the paywall — against a floor of 24.
- **No app-shell toast** — uncategorized or medal — appears anywhere in the wizard.

### Both price tiers

`SUBSCRIPTION_PRICE_TIER` picks the tier, and the tier carries the displayed price, the struck price and the Stripe lookup key as one unit, so what is shown is what is charged. Production pins `low` today and the default is `high`; both were exercised.

- `low`: €3.99/month, €23.88/year, "Save 50%", on the bank gate, the AI gate and the soft paywall.
- `high`: €8.99/month, €53.94/year, and Stripe's own checkout billed **€53.94** against the €53.94 the app quoted.

The welding is checkable without a deploy, per tier:

```
SUBSCRIPTION_PRICE_TIER=high php artisan stripe:sync-prices --dry-run
```

"Already in sync" means the Stripe price under that lookup key equals the number the app shows. **Worth running against production for both tiers before the `low` variable is removed** — in this repo's Stripe *test* account the `low` keys still hold the pre-experiment €7.80/€46.80, which is exactly the divergence the tier structure exists to prevent.

### The transfers the reveal cannot see

The figure is honest about itself now, but it is still every outgoing of the month. Separating a transfer from a purchase before anything is categorized was measured rather than guessed:

- Matching both legs — an outgoing against an equal incoming in another of the user's accounts within three days — finds **37%** of internal movements. The rest go to banks the app does not hold, or to accounts that move as a balance and have no rows to match.
- The same matcher would hide **1.6%** of genuine expense.
- `Transaction::isExpenseSide()`, the rule the dashboard and the cashflow already use, counts an uncategorized outflow as spending too. Fixing this in the reveal alone would make the reveal and the dashboard disagree.

So it belongs app-wide, with its own tests and its own decision, not as a line in this PR.

### Known, not fixed here

- The AI suggestion step could not be judged: the machine this ran on has no `GEMINI_API_KEY`, so the provider never answers and the step sits on "Reading your merchants".
- A `Select` in the shared `AccountForm` warns "uncontrolled to controlled" in the console. It is shared with Settings and predates this work.
- `/onboarding?step=ai-suggestions` deep-linked by hand on an empty account renders the gate with "It reads all **0** movements". Reachable only by editing the URL.

## Two paths this rule leaves dead

Neither is removed — there are legacy users in both states — but neither was designed for:

- **`EnsureUserIsSubscribed`'s `hard` gate is now unreachable.** "Has a bank and does not pay" can only happen to someone who had a plan and lost it, and that is `former-subscriber`.
- **`free_plan_escape_delay_hours` (3h) has nothing left to protect.** It holds the free‑plan door shut so a user is not invited to throw away a bank they just connected; there can no longer be one without a plan. Documented in config rather than deleted, because it still governs the users who got there under the old rules.

## Not switched on

`config/subscriptions.php` declares the `pay_now` / `trial` experiment but leaves `started_at` unset, so every user stays legacy and every user gets the plan defaults. Starting the split while the flow is still changing would spread the cohort across two different products. Turn it on once this has been live for two or three weeks.


victor-falcon:
## The flow, end to end, in one recording

https://github.com/whisper-money/whisper-money/raw/demo-onboarding-flow/onboarding-demo.mp4

2:49, driven against this branch. Nothing is stubbed: the checkout is a real Stripe test charge and the bank is the real Enable Banking sandbox, SCA included.

| | |
| --- | --- |
| 0:00 | The promise, then the four questions — goal, how they track today, the guess, and what happens next. |
| 0:22 | The accounts hub, and **Connect a bank** reaching the gate instead of the picker. €53.94 today, the money-back row, the disclosure row. |
| 0:40 | Stripe: **€53.94 charged**, the same figure the screen quoted. Back on *Payment received*, then straight into the bank picker it left from. |
| 1:05 | Spain → BBVA → the hand-off screen → the bank's own login and SCA → the callback. |
| 1:35 | The mapping screen, seven accounts, then the hub with their IBAN tails: `Checking · EUR · •••• 0191`. |
| 1:50 | **Add one myself**, a manual account, and a 312-row file behind it: columns, preview, progress, and the two rows it could not read. |
| 2:15 | *Reading your history*, then the reveal: last month against the guess. |
| 2:30 | The AI step, the categorizer, the first target, and the close. |

The AI step lands on *That didn't finish* and is skipped: the machine that recorded this has no `GEMINI_API_KEY`, so the provider never answers. Everything else in the recording completed for real.

Afterwards, in the database: `onboarded_at` set, 8 accounts, 533 movements, 229 categorised, subscription `active` with `trial_ends_at: null`.

PR #1016, opened 2026-09-17
feat(transactions): quick add a manual transaction from anywhere

victor-falcon:
## Why

Adding a transaction by hand is a mainstream way to use this app, not a fallback:

- **36% of users with transactions create some by hand** — 323 of 900 over 90 days.
- **24.8% of active users are manual-only** — 276 of 1,113 have no bank connection and no import at all. One in four.
- Manual rows are only 2.7% of volume (15,055 of 559,155), but the tail is long: **p90 is 46 manual entries a month**, and 117 user-months sit at 21 or more.

The problem was never reaching the form — the button already existed in two places. It was
typing 46 movements a month into a nine-field dialog.

PostHog put this segment at 51 users instead of 323, a 6x undercount, because **nothing in
the app has ever emitted a transaction-creation event**. That is why half of this PR is
instrumentation.

**Recurring transactions stay out of scope, on the data rather than to keep it small**: only
10 of 372 users (2.7%) repeat the same description+amount pair across three months or more.

## What changed

**The create dialog opens collapsed.** `EditTransactionDialog` in `mode="create"` now shows
the expense/income toggle, the amount with its currency picker, the description, and three
chips — account, date and balance. Category, labels, notes, the full account select, the full
date field and the balance checkbox sit behind **"More options"**, using the same progressive
reveal the file already had for notes. **`mode="edit"` is untouched**, and a test pins that.

- **Account chip** — the last account a manual transaction went to, remembered in
  `localStorage`; the account being viewed still wins on the account page. It is the account
  `Select` itself wearing a pill, so picking from it is one tap.
- **Date chip** — "Today" by default; it reveals the date field and flips its chevron.
- **Balance chip** — a toggle, not a selector (`aria-pressed`). It reads without colour: a
  circled € and "Updates the balance" when on, a struck icon, a dashed border and "Leaves the
  balance alone" when off. It is absent on a connected account, where the bank owns the
  balance. The expand/collapse state is deliberately not remembered — `expanded` on the event
  below is how we find out whether it should be.

**"Save and add another"** keeps the dialog open, carries the account, date and type over,
clears the amount, description and notes, and puts the focus back on the amount. That one is
for the p90.

**A global `+ Transaction` button** in the app header, right before the import button, which
is now icon-only at every width. Two sibling buttons, no dropdown, no split button, no FAB.
It loads its lists on open through a new `useTransactionDialogData` hook, shared with the
import button so there is one copy of that fetch. With no transactional account the button is
disabled with a tooltip; that is what the new `hasTransactionalAccounts` shared prop is for,
mirroring `filterTransactionalAccounts` through `AccountType::withTransactionLedger()`.

**Automation rules now say what they did.** When a rule sets the category while the form was
collapsed, the toast reads "Transaction saved · A rule categorized it as X" with "Change
category" and "Undo", instead of filing it into a category the user never saw.

**The duplicate buttons are gone** from the transactions action bar, which keeps Analysis and
Categorize. The button on the account page stays: it preloads the account being viewed, which
the header cannot.

**`transaction_created`** now fires on every manual create with `source`, `origin`,
`expanded`, `saved_and_added_another`, `account_chip_changed`, `date_chip_changed` and
`rule_applied_category` — the agreed success metrics, including chip corrections as the early
signal that "last account used" is the wrong default.

## Demo

The whole flow, desktop first and a phone at the end. Cursor and clicks are drawn in, so it
reads as somebody using it rather than a UI moving on its own.

https://github.com/user-attachments/assets/08aa4685-a93b-4fc2-9476-c1a318dca603

### The collapsed form, light and dark

The create dialog collapsed
The same dialog in dark mode

### The balance chip switched off

Dashed edge and a struck icon, so it never depends on the colour alone.

The balance chip switched off

### On a phone

The header keeps the button, without its label. The dialog is the same form,
one column: the chips wrap to two rows and the footer stacks.

The header on a phone
The collapsed form on a phone
The collapsed form on a phone, dark
The expanded form on a phone

The amount row stacks below `sm` as part of this: the toggle, the currency
picker and the number could not share 390px, and the number was rendering as
"38." before the change.

### What "More options" reveals

The create dialog expanded

### The action bar, with creating and importing gone

The transactions action bar

### Nowhere to file one yet

The disabled button and its tooltip

## Five bugs found on the way, fixed here

1. **Automation rules were switched off for every hand-typed transaction.**
   `checkAndApplyAutomationRules` returned early whenever there was no encryption key in
   storage, which is every account not on the legacy encryption. The rule engine takes a
   nullable key — it needs one to read an *encrypted account name*, not to match a rule — so
   the guard is gone. Without it the toast above would have had nothing to announce, and the
   silent categorization the brief describes was never actually running.
2. **"Save and add another" left the focus on `<body>`.** The amount is disabled while the
   save is in flight and `focus()` on a disabled input is dropped, so the next entry needed a
   click. It now waits for the render that re-enables the field.
3. **The account chip opened empty on a first visit** — nothing remembered, no account page —
   and the save came back "Account is required" after everything had been typed. It now falls
   back to the first account money can go to. That changes an existing test that asserted no
   account is auto-selected in create mode; the canvas is explicit that pre-filling is fine
   as long as the chip shows what it picked, which it does.
4. **Nothing on `/transactions` refreshed after a save.** Two causes, one
   symptom: the flag saying something had been saved was state, and the dialog
   reports the save and closes in the same batch, so the close handler still
   read it as `false`; and `router.reload()` cannot refresh that table anyway —
   its options type omits `preserveState` because it always preserves it, and
   the table copies its rows into component state on mount and never reads the
   prop again. `refreshPageAfterWrite()` visits the same URL instead, keeping
   the scroll position. The same helper fixes importing from the header, which
   had quietly lost the `onImportComplete` wiring the removed menu item had.
5. **The amount was clipped on a phone.** The type toggle, the currency picker
   and the number cannot share 390px, so "38.50" rendered as "38.". The row
   stacks below `sm` now, which is also what the canvas draws.

## Four things worth a decision, not fixed here

1. **`accounts`, `categories`, `banks`, `labels` and `automationRules` are already shared on
   every Inertia response.** `HandleInertiaRequests::userCollectionProps()` returns them as
   plain closures, and Inertia only skips `OptionalProp`/`DeferProp`, so the docblock calling
   them "deferred props" is wrong. The on-demand `fetch` this PR reuses therefore re-fetches
   data the page already carries. The brief asked for the fetch, so that is what is here, but
   either those props should become `Inertia::optional()` or the fetch should go.
2. **The create `EditTransactionDialog` on `/transactions` is now unreachable.** The brief
   said to keep it and remove only the button that opened it, so it is still mounted, but
   nothing sets `createDialogOpen` any more. The same is true of the one in
   `transaction-list.tsx`, whose actions menu was already dead (`showActionsMenu={false}` at
   all three call sites). Say the word and both come out.
3. **The account page's dialog gets "Undo" but not "Change category"**, because that action
   needs an `onRequestEdit` and wiring one there means reworking a file the brief ruled out.
4. **New strings went into `lang/es.json` and `lang/fr.json` only.** `lang/en.json` holds a
   single deliberate override out of 3,480 keys — English source strings fall through by
   design, so identity entries there would be noise. Happy to add them if the rule is meant
   literally.

## Tests

New and updated coverage for: the collapsed form and what "More options" reveals, edit mode
staying whole, the account chip reading the remembered account, the date chip, the balance
chip and what it persists, "Save and add another", the rule-categorized toast with its undo,
the `transaction_created` payload, the header button's disabled state, the action bar losing
both buttons, `hasTransactionalAccounts` (empty, positive, archived and balance-only), and
`labels` on the import-data endpoint.


victor-falcon:
Mobile screenshots added to the description, under **On a phone**.

